PR #25, opened 2015-12-09
Extend .gitignore with common IDE pattern and more

sirthias:


jenshalm:
Thanks for the editions. But I think at least three of the removals need to stay. I still see these files in Scala IDE (e.g. `.cache-main`, `.cache-test` and `*.sc` for worksheets).

sirthias:
Sorry, I didn't mean to remove stuff.
I simply copied over a "trusted" `.gitignore` from another project, which apparently wasn't that trustworthy. :)
So at least now we have something even better.

jenshalm:
Cool, thx!

PR #37, opened 2017-02-13
Scala 2.12

mscharley:
An attempt at updating to Scala 2.12. This compiles and passes the current test suite though there's lots of warnings and deprecation notices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Open Source under the Apache 2.0 License.
 Getting Started
 ---------------
 
-The main artifact is published to Maven Central for Scala 2.10 and 2.11.
+The main artifact is published to Maven Central for Scala 2.10, 2.11 and 2.12.
 
 The sbt plugin is published to the sbt plugin repository for sbt 0.13.x.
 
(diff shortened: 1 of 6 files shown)


mscharley:
Is there any chance of this getting into a release any time soon? If not then we may need to look at migrating to another library.

zifeo:
@jenshalm I would be great if you can merge this 👍

jenshalm:
I will merge this week, just want to find a minute to investigate why this strange TryOps hack is needed.

Releasing will take a bit more time, probably some time in March.

mscharley:
Hi Jena, thanks for the reply. The TryOps things is because 2.12 introduced their own version of toEither which conflicts with yours. The other option I considered was updating your version to match the signature of the new one which `def toEither(): Either[Throwable,T}` but this was the path of least resistance to a working build. I'm more than happy to make this update if this is the way you'd rather go with this.

jenshalm:
Ah, I actually forgot I had defined my own implicit for `toEither`. I think I merge this as is for now, and maybe later rename my own implicit.

jenshalm:
Many thanks for this PR, btw. :-)

mscharley:
No problem at all. Was quick to do since there really weren't any issues and Laika is the last library we're waiting on to update one of our applications to 2.12.

mscharley:
@jenshalm when is this likely to get into a release?

jenshalm:
I need one evening for some minor polishing and one more evening for the actual release which sadly is not an automated process. This week it's too warm and sunny :-) but chances are high this changes next week, I live in London after all...

jenshalm:
The release is out now, I took the time to fix a few issues first which had been open for too long.

PR #174, opened 2021-01-19
Add highlight support for YAML

note:


core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+
+    (attributeName ~ valueLiteral).concat
+  }
+
+  private val key: CodeSpanParser = {
+    CodeSpanParser((keyName ~ oneOf(':') ~ anyOf(' ','\t') ~ oneOf('\n')).map { r =>
+      List(r._1._1._1, CodeSpan(r._1._1._2 + r._1._2 + r._2))
+    })
```
jenshalm, line 63: This would need a similar refinement as the `keyAndValue` parser for optional spaces before the colon.

core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+  /** The names of the language (and its optional aliases) as used in text markup */
+  override def language: NonEmptyList[String] = NonEmptyList.of("yaml", "yml")
+
+  private val keyName = Identifier.alphaNum.withIdStartChars('_', '-', '.').withCategory(CodeCategory.AttributeName)
+
+  private val valueLiteral = {
+    val doubleQuotedStrLiteral = StringLiteral.singleLine('"')
+    val singleQuotedStrLiteral = StringLiteral.singleLine('\'')
+    val number = NumberLiteral.decimalFloat | NumberLiteral.decimalInt
+    val boolean = (literal("true") | literal("false")).map(b => Seq(CodeSpan(b, CodeCategory.BooleanLiteral)))
```
jenshalm, line 41: YAML has a long list of valid boolean literals:
```scala
val boolean = Keywords(CodeCategory.BooleanLiteral)(
  "true","True","TRUE","false","False","FALSE","on","On","ON","off","Off","OFF","yes","Yes","YES","no","No","NO"
)
```

core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+object YAMLSyntax extends SyntaxHighlighter {
+  /** The names of the language (and its optional aliases) as used in text markup */
+  override def language: NonEmptyList[String] = NonEmptyList.of("yaml", "yml")
+
+  private val keyName = Identifier.alphaNum.withIdStartChars('_', '-', '.').withCategory(CodeCategory.AttributeName)
+
+  private val valueLiteral = {
+    val doubleQuotedStrLiteral = StringLiteral.singleLine('"')
+    val singleQuotedStrLiteral = StringLiteral.singleLine('\'')
+    val number = NumberLiteral.decimalFloat | NumberLiteral.decimalInt
```
jenshalm, line 40: Octal and hex integers are also supported (and quick to add here).

core/shared/src/test/scala/laika/parse/code/languages/LanguageSpec.scala
```diff
@@ -642,6 +642,55 @@ class LanguageSpec extends AnyWordSpec with Matchers {
...
+      val nl = other("\n")
+      val nl2 = other("\n  ")
+      val colon = other(": ")
+      parse(input) shouldBe result("yaml",
+        comment("# Source: mychart/templates/configmaps.yaml\n"),
+        attrName("apiVersion"), colon, string("v1"), nl,
+        attrName("kind"), colon, string("'ConfigMap'"), nl,
+        attrName("metadata"), other(":\n  "),
+        // It would be more accurate if the space of `other(" ")` was part of its preceding string("port_overriden")
+        // but in the end it does not matter how space will be colored
```
jenshalm, line 679: I'd suggest to remove this comment as it is misleading. In fact, the way you implemented it is exactly correct for two reasons: a) comparable highlighters in Laika behave in the same way, b) a YAML parser would not include the whitespace before the `#` in the result.

core/shared/src/test/scala/laika/parse/code/languages/LanguageSpec.scala
```diff
@@ -642,6 +642,55 @@ class LanguageSpec extends AnyWordSpec with Matchers {
...
+          |  timeout: 23.56
+          |  null_value: null
+          |  boolean: true
+          |a_sequence:
+          |  - Item 1
+          |  - 123
+          |  - key: value
+          |```
+        """.stripMargin
+
```
jenshalm, line 669: It would be good to vary the syntax a bit with optional spaces before and after the colon to test the suggested changes.

jenshalm:
There are a few more comments this time, but I think they are all fairly simple changes.
Otherwise this is good to go.

I'll look at the final PR for Dhall in the coming days, but that might take a tad longer as it seems to be a slightly bigger piece of work and don't know Dhall very well.

core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  private val keyAndValue = {
+    val attributeName = {
+      // https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html#yaml-basics:
+      // the colon must be followed by a space
+      (keyName ~ oneOf(':') ~ oneOf(' ','\t')).map { r =>
+        List(r._1._1, CodeSpan(r._1._2 + r._2))
+      }
+    }
```
jenshalm, line 55: The actual syntax for key+value is much less restrictive. The precise grammar is zero or more spaces before the colon and one of more after it.

I'd suggest the following alternative (untested, but probably fine):
```scala
private val keyAndValue = {
  val separator = (anyOf(' ','\t') ~ ":" ~ someOf(' ','\t')).source.map(s => List(CodeSpan(s)))
  (keyName ~ separator ~ valueLiteral).concat
}
```

core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
note: It depends on YAML spec version, in 1.2 it's only 2 literals: https://github.com/go-yaml/yaml/issues/214

Anyway, I think I will go with 1.1 as you suggested, probably it's makes more sense to be more liberal for highlighting

jenshalm: I wasn't aware of that change, I looked at an unversioned grammar document on yaml.org and tested a few common parsers like `pyyaml` which apparently never got updated for YAML 1.2.

I don't have a strong opinion either way, but if the 1.1 booleans are still in wide-spread use it might indeed make sense to support them.

core/shared/src/main/scala/laika/parse/code/languages/YAMLSyntax.scala
note: Thanks, it worked perfectly

note:
@jenshalm I addressed your comments, it's ready for re-review

PR #175, opened 2021-01-24
Syntax highlighting for Haskell

note:


core/shared/src/main/scala/laika/parse/code/languages/HaskellSyntax.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  val stringLiteral =
+    StringLiteral.singleLine('"').embed(charEscapes)
+
+  val numberLiteral =
+    NumberLiteral.hex ++ NumberLiteral.octal ++ NumberLiteral.decimalFloat ++ NumberLiteral.decimalInt
+
+  val identifiers = IdParser(lowerAlpha.add('_').add('\''), digit ++ upperAlpha)
+
+  val types = IdParser(upperAlpha.add('_').add('\''), digit ++ lowerAlpha).withCategory(CodeCategory.TypeName)
```
jenshalm, line 32: Same question as for `identifiers`.

core/shared/src/main/scala/laika/parse/code/languages/HaskellSyntax.scala
```diff
@@ -0,0 +1,43 @@
+package laika.parse.code.languages
```
jenshalm, line 1: Like with the other PR's, it would be better to add the license header to all new class files (and an author tag to the class scaladoc, if you wish).

core/shared/src/main/scala/laika/parse/code/languages/HaskellSyntax.scala
```diff
@@ -0,0 +1,43 @@
...
+  val keywords =
+    Keywords("as", "case", "data", "data family", "data instance", "default", "deriving",
+      "deriving instance", "do", "else", "forall", "foreign", "hiding", "if", "import", "in",
+      "infix", "infixl", "infixr", "instance", "let", "mdo", "module", "newtype", "of",
+      "proc", "qualified", "rec", "then", "type", "type family", "type instance", "where"
+    )
+
+  val stringLiteral =
+    StringLiteral.singleLine('"').embed(charEscapes)
+
```
jenshalm, line 26: Adding `charLiteral` would be nice. You are right that none of these syntax definitions are supposed to cover 100%, but char literals are supported in all other languages, so it would make it more consistent.

jenshalm:
Only these four comments from my side, otherwise good to be merged.

core/shared/src/main/scala/laika/parse/code/languages/HaskellSyntax.scala
```diff
@@ -0,0 +1,43 @@
...
+  val numberLiteral =
+    NumberLiteral.hex ++ NumberLiteral.octal ++ NumberLiteral.decimalFloat ++ NumberLiteral.decimalInt
+
+  val identifiers = IdParser(lowerAlpha.add('_').add('\''), digit ++ upperAlpha)
```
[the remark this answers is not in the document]
note: Good catch, thanks!

note:
Thanks for your comments, I addressed them

jenshalm:
Excellent, thanks a lot.

PR #176, opened 2021-01-28
Syntax highlighting for Alloy

note:


jenshalm:
Hello and sorry for the delayed review. As you probably noticed from the low number of commits this winter I took a short hiatus from Open Source work. I'll now resume with the Laika road map, and your PRs could be part of the 0.17.1 release which I plan to do in March.

I'll start with Alloy as I know nothing about it, so have to entirely rely on your judgement of which things to include instead of a proper review.
In that respect this is ready to go, apart from the missing license header in the new source file.
You can copy it from any other source file. It contains the line "Copyright the original author or authors", so you can also add a scaladoc comment on the main class of that file and add an `@author` tag with your name, unless you insist on remaining anonymous. I'm not a lawyer, but I think you adding the license yourself is more sound than me slapping it onto it retroactively.

note:
Hi Jens! That's totally understandable you took a break from OSS.

I've added the said headers and annotations to this PR. I will add it to others later

PR #177, opened 2021-02-18
Syntax highlighting for Dhall

note:
Hi @jenshalm, sorry for bombarding you with so many PRs. I have a single branch with changes from all 4 PRs. If you prefer to merge things in one go, just let me know.

None of those PRs are perfect in sense they do not support grammars in 100%. That's a fortunate aspects with syntax highlighters that it's purely additive excercise, even having 20% is better than 0% :)

### DIFF
--- a/core/shared/src/main/scala/laika/parse/code/SyntaxHighlighting.scala
+++ b/core/shared/src/main/scala/laika/parse/code/SyntaxHighlighting.scala
@@ -58,6 +58,7 @@ case object SyntaxHighlighting extends ExtensionBundle { self =>
       HOCONSyntax,
       XMLSyntax,
       YAMLSyntax,
+      DhallSyntax,
       SQLSyntax,
       EBNFSyntax,
       MarkdownSyntax,
--- a/core/shared/src/main/scala/laika/parse/code/languages/DhallSyntax.scala
+++ b/core/shared/src/main/scala/laika/parse/code/languages/DhallSyntax.scala
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package laika.parse.code.languages
+
+import cats.data.NonEmptyList
+import cats.implicits._
+import laika.ast.~
+import laika.ast.CodeSpan
+import laika.bundle.SyntaxHighlighter
+import laika.parse.code.common.NumberLiteral.NumericParser
+import laika.parse.code.common._
+import laika.parse.code.{CodeCategory, CodeSpanParser}
+import laika.parse.implicits._
+import laika.parse.text.{CharGroup, PrefixedParser}
+import laika.parse.text.TextParsers._
+
+/**
+ * https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf
+ * @author Michał Sitko
+ */
+object DhallSyntax extends SyntaxHighlighter {
+  /** The names of the language (and its optional aliases) as used in text markup */
+  override def language: NonEmptyList[String] = NonEmptyList.of("dhall")
+
+  val comment =
+    Comment.singleLine("--") ++ Comment.multiLine("{-", "-}")
+
+  val keywords = Keywords(
+    "if", "then", "else", "let", "in",
+    "using", "missing", "assert", "as",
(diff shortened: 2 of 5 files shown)


note:
I also included a tiny fix for a bug in JS `PlatformDateFormatImpl`. I spotted the bug because CI tests failed with:

```
[info] - should use the default pattern when no pattern is specified *** FAILED ***
[info]   RootElement - Blocks: 1
[info]   . Paragraph - Spans: 1
[info]   . . Text - 'Some 2/25/2021, 6:44:55 AM'
[info]    was not equal to 
[info]   RootElement - Blocks: 1
[info]   . Paragraph - Spans: 1
[info]   . . Text - 'Some 2/25/2021, 6:44:54 AM' (StandardSpanDirectivesSpec.scala:153)
[info]   Analysis:
[info]   RootElement(content: List(0: 
[info] Paragraph - Spans: 1
[info] . Text - 'Some 2/25/2021, 6:44:55 AM'
[info]  -> 
[info] Paragraph - Spans: 1
[info] . Text - 'Some 2/25/2021, 6:44:54 AM'
[info] ))
```

jenshalm:
I've now looked at this final PR, too, with the limitation of being someone who never used Dhall in practice, which means that there are some subtle details where I'll just trust you that you knew what you were doing. :-)

I've just two little requests for this one:

* Could you add support for boolean literals, or was there a specific reason you excluded them?

* Could you please revert/exclude all the work around your date formatter fixes?

The reasons for reverting the date fix are twofold:

* The flakiness of the test is actually not directly related to the issue with the date directive and more coincidentally fixed by your changes. I had removed the flakiness already a week or two ago by enhancing just the test itself (zero-ing the seconds before comparing the result). It previously relied on two constructed dates always being in the same second when rounded, which always worked on my machine, but might fail on slower CI nodes as you noticed.

* The issue with the date directive needs to be addressed differently. You fixed it in a generic utility which will also change the output of the built-in HOCON renderer for example, which is an undesired change. I'll create a separate PR that changes just the implementation of the directive and ping you so that you can have a look if you want.

jenshalm:
Ah, and another thing just FYI: I was puzzled why you manually added support for a sign to the number parser until I realized that for reasons unknown I never added that to the built-in parsers. That was an accidental omission. This means we can keep what you did for now, but for the upcoming release I'll add general support for signs which most/all languages need and then remove your explicit support from the Dhall syntax again.

jenshalm:
@note : the PR for an improvement of the rst date directive when run with with Scala.js - #180

note:
> Could you please revert/exclude all the work around your date formatter fixes?

Sure, will do. 

The reason I included it and regarded it as a fix was that it was adding time components to `date` directive. I looked up for a spec and back then it appeared to me that there should be no time at all. Now, looking at https://docutils.sourceforge.io/docs/ref/rst/directives.html#date I see that it says `(...) but time fields can also be used`. That makes your fix more aligned it specs than main.

I will respond to other questions later in the day

jenshalm:
Yes, there are some subtleties around how to support the date directive for Scala.js as it cannot be done 100% properly with current limitations of the Date support in Scala.js.

But to clarify, the main reason we could not keep your fix was not related to the date directive itself, but that your fix broke the HOCON renderer, where the generic `PlatformDateFormat` is also used (https://github.com/planet42/Laika/blob/master/core/shared/src/main/scala/laika/config/ConfigEncoder.scala#L61). That usage is completely unrelated and would have been affected by the change. When rendering HOCON there is no reason to exclude the time component.

note:
> Could you add support for boolean literals, or was there a specific reason you excluded them?

There's no notion of "bool-literal" in Dhall's grammar: https://github.com/dhall-lang/dhall-lang/blob/master/standard/dhall.abnf. Instead, it's defined as `builtin` which contains a bunch of other things like `Natural-fold` and other built-in types.

I did a bit of archeology on topic whether `True` should be treated as literal and here are relevant discussions: https://github.com/dhall-lang/dhall-lang/pull/490, https://github.com/dhall-lang/dhall-lang/pull/484, https://github.com/dhall-lang/dhall-lang/pull/490. That said, I think that in scope of syntax highlighter most of those subtleties do not matter.

All in all, I don't have a strong opinion, if there are no additional considerations I would leave it as is.

note:
> I was puzzled why you manually added support for a sign to the number parser until I realized that for reasons unknown I never added that to the built-in parsers.

Curiously, in Dhall plus sign is required for `Integer` and illegal for `Natural`. I just added a test for explicit plus sign

> This means we can keep what you did for now, but for the upcoming release I'll add general support for signs which most/all languages need and then remove your explicit support from the Dhall syntax again.

Sounds good

note:
I've made most of improvements, let me know if something need to be worked on

jenshalm:
Thanks, but the date fix is not fully reverted, there are still changes in `PlatformDateFormatImpl` and `ConfigCodecSpec`. The former is just whitespace, but the latter is significant as it does no longer test for the time component being preserved in a decode/encode round trip.

note:
Fixed

jenshalm:
Great, I'll merge this now and then I'll probably prepare a release (0.17.1) within the next 10 days or so which will contain your PRs amongst a range of other things. I might look into adding support for escapes in String literals for Dhall as most other languages support it in which case I'll do a PR and ping you in case you want to have a look. You seem to know more about Dhall than I do.

note:
Sure, feel free to reach me out

PR #180, opened 2021-03-15
Improve rst date directive for Scala.js

jenshalm:
This works around the limitation of JS Dates not supporting an explicit pattern in toLocaleString by making an educated guess whether the result is expected to contain a time component based on the user-specified pattern.
  
A proper way to handle this would be to parse the pattern and translate it to a JavaScript options object, but this is currently considered beyond the scope of Laika.
Only very few users are affected by this limitation, as the method is currently not public API. 
The only place where it is indirectly exposed to users is the date directive for reStructuredText, which allows to pass a pattern. As a consequence, using this directive with Scala.js is currently not fully supported.


note:
Looks good to me

PR #181, opened 2021-03-17
Support escapes and interpolation in Dhall syntax highlighter

jenshalm:
This is what I extracted from the documentation and ABNF. It is fairly similar to Scala and reuses existing building blocks apart from the braced unicode escapes for which no matching built-in construct exists.

(FYI @note).


note:
Looks good to me. Thanks for taking care of that

PR #198, opened 2021-05-16
Add Scala 3 keywords

note:
I have not added [soft keywords](https://dotty.epfl.ch/docs/reference/soft-modifier.html) as they seem more involving


jenshalm:
Thanks for this PR.

However, there is already a `DottySyntax` implementation which also includes some of the soft keywords (although I just noticed that some syntax may have changed in the meantime): https://github.com/planet42/Laika/blob/master/core/shared/src/main/scala/laika/parse/code/languages/DottySyntax.scala

The question would be more whether we just change the binding now, so that when you add a scala code block, you automatically get dotty syntax, even when you write Scala 2. It could also be configurable, but that's probably overkill.

note:
I completely missed that 🤦 

Annotating snippets with `dotty` works for me, I guess for some time I will switch between scala 2 and scala 3 so it's not a problem for me

PR #229, opened 2021-10-22
preview server: add host field

i10416:
Hello! I tried Laika and found out that I need to set host 0.0.0.0 instead of `localhost` when preview server is running in WSL, so I added host field to `LaikaConfig` and `LaikaPreviewConfig` so that developers can set host from `build.sbt`.


jenshalm:
This is now ready to merge, as the main branch will be for 0.19 development now. However, it went out of sync with the main branch in the meantime and I resolved the conflicts incorrectly and cannot do any further changes on your branch. 

In https://github.com/planet42/Laika/pull/229/files#diff-a9449928d5aafc3cac5a0c2ad17ef44db0123b5af201382eab8066fdbf3de3b0R83 the `ctx` parameter needs to get removed, after that the tests should be green.

i10416:
Thank you! I fixed compile error.

jenshalm:
Thank you! Merging now...

PR #240, opened 2021-11-05
add location

i10416:
Improve tests so that  they can show the correct location where test fails.

see https://scalameta.org/munit/docs/tests.html#declare-tests-inside-a-helper-function


jenshalm:
Thanks a lot, that's a great find! I was actually aware of the need of this parameter and added it in a few places, but apparently you found dozens of places where I did not. :-) It was a mass migration of 2K+ tests, and a few things were probably done a bit too quickly, there are further cleanup tasks on my todo list.

Issue #137, opened 2020-04-20
Implement @:siblingNavigation directive


jenshalm:

jenshalm:
Superseded by #245.

PR #245, opened 2022-01-15
add prevDoc and nextDoc template directives

i10416:
rel #137

cursor.previousDocument.(absolutePath|relativePath) is not useful when
it comes to adding links to prev/next documents because it returns
not resolved path(e.g. path/to/index.html) but unresolved path(e.g. path/to/README.md)

Besides, it is not possible to eval values in `path` directive.  

I tried path directive with Substitution Variables, but it results in error.

```
@:path(cursor.previousDocument.absolutePath) 
@:path(${cursor.previousDocument.absolutePath}) 
```

`prevDoc` and `nextDoc` ,for example, enables something like this.

image

### DIFF
--- a/core/shared/src/main/scala/laika/directive/std/StandardDirectives.scala
+++ b/core/shared/src/main/scala/laika/directive/std/StandardDirectives.scala
@@ -118,6 +118,23 @@ object StandardDirectives extends DirectiveRegistry {
     }
   }
 
+  lazy val prevDoc: Templates.Directive = Templates.create("prevDoc"){
+    import Templates.dsl._
+    cursor.map { cursor =>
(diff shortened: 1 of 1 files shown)


jenshalm:
While this PR is definitely a possible solution to the problem you describe, it has two downsides:

1) It leaves the underlying issue you described unresolved (that the `previousDocument` and `nextDocument` variables are not very useful in templating). People might still be tempted to use them and run into the same issue as you did.
2) It is quite prescriptive, e.g. it would not allow to use icons as part of the links, whereas the variables give you all the flexibility to define your UI.

As part of `laika-core` I'd therefore prefer to use this alternative PR (#262) instead, which addresses issue 1).
For a more prescriptive UI element there is still the option to add it to the Helium theme later, which also allows for convenient icon references for example.

PR #254, opened 2022-02-08
core: add dart syntax highlight

i10416:


i10416:
Hi, I added Dart syntax support. Please check when you have time.

Thank you in advance.

i10416:
@jenshalm

jenshalm:
Apologies for the delay, I finally found the time to look at this PR. It took some time as I have never used Dart before, which means I had to find and study some grammar documents for this review.

As a general rule for syntax highlighters it is fine to only recognise a subset of the syntax, as this is solely meant to be used as a lightweight tool for highlighting languages in documentation, not as a basis for tools like IDEs, as there are existing solutions which are much more complete and mature for this purpose.

The features which are recognised, however, need to be as accurate as possible. And from the documentation I found (for example this grammar document which is extracted from the spec: https://github.com/chalin/dart-spec-and-grammar/blob/master/doc/grammar.pdf) it seems that there are significant differences between this syntax implementation and the spec. I'll list the things I found below.

**String Literals**

* Multiline strings seem to be delimited by either triple-quote or triple-doublequote, not by the backtick you have chosen.
* Substitutions seem to be supported in single-line strings, too. This PR only adds them to multiline strings.
* It would be good to also support raw strings, as they are interpreted very differently (no escapes, no substitutions).

**Number Literals**

* This PR adds underscores as delimiters, but they appear to be a plan for the future, but not yet supported: https://github.com/dart-lang/language/issues/2
* This PR adds octal and binary literals, which do not appear in the spec.
* This PR adds suffixes for long, float and bigint, which also do not appear in the spec. 

**Builtin Types**

* `float`, `long` and `byte` do not appear in the spec.

**Builtin Identifiers**

* There are some identifiers in this list which are not supported in this PR: https://github.com/dart-lang/language/blob/master/specification/dartLangSpec.tex#L16549 
    As I don't know Dart I am not sure whether they should all be supported. I usually avoid adding soft keywords, as they tend to often show up as false positives. Just have a quick look whether any of these should be added as many of them are missing in this implementation.

These issues would need to be addressed for this to be merged. There is no need to rush though, as there is no release planned for the coming weeks. If you think that some of the listed issues are actually correct, it would be good if you can point me to a spec document that describes the feature.

jenshalm (changes requested):
See list of potential issues in the previous comment.

i10416:
> The features which are recognised, however, need to be as accurate as possible. ...

Thank you for reviewing. I assumed without serious consideration that Dart syntax is `subset` of JavaScript + Java (actually, it feel like writing JS one time, and Java another to write Dart code😓) and shamefully copy-paste some parts from those 🙇

i10416:
- [x] Builtin Types: remove float, long and byte

i10416:
AFAIK,
- as for `covariant`, `deffered`, `external`, `interface` and `set`,  they are rarely used in dart codebase
- people often prefer method declaration like `void f(){...}` or lambda to `function` keyword
- `export`, `part`, `part of` ,`show` , `hide` and `library` are mainly used for module management

> tend to often show up as false positives

If you are not happy with false positives, I think they will not make much difference.

On the other hand, I forgot adding `get`, `typedef` and `operator`.

> I am not sure whether they should all be supported

As Dart is object oriented language, `abstract`, `class`,  `extends`, `implement`, `super` and `with` are often used. In addition, `final` and `const` should be included because Dart linter often suggests them instead of `var` or type annotation. `required` keyword is often used in combination with null safety feature.

I think `this` keyword should be included too because it carries contextual information though Dart linter often suggests removing redundant `this`. 

`switch` and `case` also should be included as there is a plan(perhaps proposal?) to add rich enum like sealed trait.

i10416:
todo
- [ ] tests for multiline strings
- [ ] raw string
- [ ] correct number representation

jenshalm:
Sorry, regarding "false positives" for keywords I wasn't really clear. The key difference I forgot to explain is whether something is a soft modifier/keyword or a hard modifier/keyword.

Soft keywords are only recognised in certain positions, but can still be used as variable or method names in other positions. These are the keywords I usually avoid to add to the highlighter. Since Laika does not do full syntax analysis, we usually have no easy way to determine whether the use is in keyword position or not. And having a method name highlighted as a keyword looks confusing. Example for soft modifiers in Scala 3 are documented here: https://docs.scala-lang.org/scala3/reference/soft-modifier.html

Hard keywords on the other hand are always safe to support. In this case it does not really matter if they are rarely used or not. You cannot get false positives as they are illegal identifiers for methods or variables. So it's always good to include all hard keywords and modifiers.

If you cannot easily find documentation for what are hard or soft keywords, a quick check with the compiler is usually sufficient: if the compiler does not allow you to use the keyword as a variable name, for example, it's safe to add it to the highlighter.

i10416:
Thank you for clarification. I misunderstood what you mean by soft keyword.

> In this case it does not really matter whether they are rarely used or not

Okey, I put some keywords back.

i10416:
@jenshalm ping

jenshalm (approved):

jenshalm:
Thanks a lot, looking good now!

Issue #231, opened 2021-10-25
Consider using `fs2-io` for `laika-io`


armanbilge:
Just a suggestion :)

The `fs2-io` module provides `Files[F]` for dealing with file I/O: https://github.com/typelevel/fs2/blob/main/io/shared/src/main/scala/fs2/io/file/Files.scala

IMHO it's worth considering for two reasons:
1. I took a quick look through your sources and I think there's a few places where you should be using `Sync.blocking` instead of `Sync.delay` b/c otherwise you are running blocking I/O on the CE3 primary threadpool which can lead to starvation etc. and other sorts of problems. If it's helpful I can try and find a reference about this. If you use fs2, this is all handled correctly for you.
2. `fs2-io` actually lets you cross-compile your code to Scala.js and run on Node.js. This is interesting, if you want to make a command-line version of Laika because Node.js has _much_ faster startup time (essentially instant) than JVM.

jenshalm:
Closing this now as superseded by #281 which is the new umbrella ticket for tracking all PRs of this migration.

Thanks for the initial nudge! It looks like this is going to be a significant improvement.

PR #257, opened 2022-02-20
migrate to fs2-io

i10416:
Hello.

In this PR, I replaced java.io stuffs with fs2.io.Files and fs2.io.Path.

I tried to keep signature as possible except for Sync/Async, but I think some API changes are needed. 

It makes sense to use fs2 library for io module despite additional dependencies because fs2 helps cross platform io operations as mentioned in #231. It will be great if laika-io runs on js too.

### DIFF
--- a/io/src/main/scala/laika/io/model/Input.scala
+++ b/io/src/main/scala/laika/io/model/Input.scala
@@ -31,10 +31,14 @@ import laika.io.runtime.{DirectoryScanner, InputRuntime}
 import laika.rewrite.nav.TargetFormats
 
 import scala.io.Codec
+import cats.effect.kernel.Async
+import java.nio.channels.Pipe
+import fs2.io.file.Files
 
 sealed trait InputReader extends Product with Serializable
 case class PureReader(input: String) extends InputReader
 case class StreamReader(input: Reader, sizeHint: Int) extends InputReader
+case class SourceReader[F[_]:Async](stream:fs2.Stream[F,String]) extends InputReader
 
 /** A binary input stream and its virtual path within the input tree.
   */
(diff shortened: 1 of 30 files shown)


i10416:
Btw, is there any reason not to use formatter?

i10416:
dep #258

armanbilge:
Wow, nice, thanks for working on this! I'm happy to help review this one :)

io/src/main/scala/laika/io/model/Input.scala
```diff
@@ -78,6 +82,12 @@ object TextInput {
     TextInput[F](path, docType, InputRuntime.textFileResource[F](file, codec).map(StreamReader(_, file.length.toInt)), Some(file))
   def fromStream[F[_]: Sync] (path: Path, docType: TextDocumentType, stream: F[InputStream], codec: Codec, autoClose: Boolean): TextInput[F] =
```
armanbilge, line 83: Btw, I think all uses of `F[InputStream]` should actually be replaced with `fs2.Stream[F, Byte]`. It is a referentially transparent, pure-FP, and `fs2.io` includes methods to get `fs2.Stream` from classpath resources etc. while correctly handling blocking.

i10416: Yes, I feel the same. However, I think we should be careful because  i/o stream is  such a fundamental component for io module that it will propagate changes.

i10416:
Thank you @armanbilge for taking time to review this PR. I will tweak this pr in a few days.

io/src/main/scala/laika/io/model/Input.scala
jenshalm: Unfortunately I have not yet looked at this PR in detail, but as a general tendency, I feel like while it would make sense to standardise on `fs2.Stream` for any *internal* API wherever feasible, I think I'd still like to offer `InputStream` in the public APIs, since it is such a common type and this lib is also intended for an audience that does not know that much about Typelevel libs. The implementation of such a method could then immediately translate down to fs2 wherever possible.

armanbilge: > I think I'd still like to offer `InputStream` in the public APIs, since it is such a common type and this lib is also intended for an audience that does not know that much about Typelevel libs.

:100: I definitely agree with this sentiment, with regard to accessibility.

However, in this case I would argue that signatures `F[InputStream]` are actually more likely to be abused by the uninformed user than `fs2.Stream[F, Byte]`.

A user who doesn't know much about Cats Effect could very plausibly write `F.pure` where they meant `F.delay` or even worse `F.blocking` and cause all sorts of problems for Laika/Cats Effect which make many important assumptions about the semantics of code suspended in `F[_]` (for example, at which point side-effects vs `pure` are evaluated and on which threadpool).

On the other hand, It seems reasonable to add scaladocs to point a user to methods such as `fs2.io.readClassResource("myResource")` to get an `fs2.Stream[F, Byte]`, which I think are easy to use, hard to abuse, and do the right thing out-of-the-box.

To offer an accessible "unsafe" API, FWIW I think you should do away with `F[_]` entirely which I suspect is confusing for this group of users as well. This API can be then implemented in an imperative style and take the appropriate measures not to violate the assumptions of the Cats Effect runtime.

jenshalm: I was actually thinking about the specific use case of interop with existing tools that provide an `InputStream`, not the simpler scenarios of classpath or file resources. Removing the existing `fromStream` method does not make anything safer here, since it just moves the problem to `fs2.io.readInputStream` which also expects an `F[InputStream]`. The way I see it it's just more code to write for the user and the same kind of trap to fall into.

My preference would be not to remove any existing API, but instead only add `fromStream(fs2.io.Stream)` and then use fs2 for most of the internal API. Which means to also keep `fromFile(String)` and `fromClasspathResource(String)`.

jenshalm: But it's a good point to improve scaladoc for users not familiar with cats-effect. From working with junior devs in my day job I am aware that confusing `F.delay`, `F.pure` and `F.blocking` is amongst the no1 sources of bugs when using effects (to some degree even for seniors, to my surprise).

i10416:
Thank you for feedback to this PR and proposing better gradual migration PRs!

PR #275, opened 2022-04-04
avoid naive seq size comparison for performance

i10416:
`lengthCompare` is usually more performant than combination of `length` and compare
operator.

### DIFF
--- a/core/shared/src/main/scala/laika/ast/Cursor.scala
+++ b/core/shared/src/main/scala/laika/ast/Cursor.scala
@@ -300,7 +300,7 @@ case class DocumentCursor (target: Document,
 
     /** The next document in a flattened tree view or None if this is the cursor points to the last document.
       */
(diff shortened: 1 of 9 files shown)


jenshalm:
Thank you for looking into this, but I am inclined to close this, as the PR does not seem to contain what the headline claims 🙂 

Most of the calls to `size` or `length` in this PR are *not* on an abstract `Seq`, but on types like `String`, `Array`, `Vector` or `ArrayBuffer` where there is no overhead in calling `size`. I did not check every single change as I've already found too many places where I see it as a change for the worse (reducing readability and simplicity of the code without any meaningful performance gain).

Feel free to reopen a smaller, more focussed PR where the call is indeed on `Seq` or `List` *and* the code location suggests that those structures could be larger.

i10416:
Thanks.

PR #282, opened 2022-04-29
use fs2 for TextInput, TextOutput, BinaryInput

jenshalm:
This is the main migration step.
    
Adjusting `TextInput` and `TextOutput` is straightforward due to the simplification in the first step.
    
`BinaryInput` will receive a breaking change by modelling the input as `fs2.Stream[F, Byte]`. This is also "public-internal" API with little impact on end users.

`BinaryOutput` is the unfortunate omission here: it cannot be migrated to fs2. The reason is that the model is currently solely used for binary output formats like PDF and EPUB which require the use of Java APIs that would not know how to handle an `fs2.Pipe`. This leaves an odd inconsistency between binary input and output, but half a migration is still better than none. This also has an impact on potential, future support for Scala.js in `laika-io` which is discussed further below.

The types `InputRuntime`, `OutputRuntime` and `CopyRuntime` in the `laika.io.runtime` package can be removed entirely. Now that fs2 is the IO runtime they are not needed and the little integration and plumbing that is required stays in the companions of the input and output model. Additional benefit is that the amount of code to maintain in `laika-io` shrinks by 200 lines of code.

Another integration challenge is the `FopResourceResolver` for Apache FOP in the PDF renderer, which provides font, image and other resources to FOP. Here we need to convert an `fs2.Stream` back to a Java `InputStream` to implement a synchronous Java callback. This is mildly inefficient, but given that PDF has by far the slowest renderer in Laika anyway, the difference should not be noticeable.

Final oddity is the `ContainerWriter` for EPUB output which now talks to a `ZipOutputStream` and an `fs2.Pipe` pointing to the same underlying `OutputStream` to fulfil differing integration requirements. At least in the test suites that works fine.

Some further manual testing is required before this can be merged, as some functionality cannot be easily checked by Laika's test suites:

- [ ] visual checks for EPUB output
- [ ] visual checks for PDF output
- [ ] working with the preview server


io/src/main/scala/laika/io/model/Input.scala
```diff
@@ -71,17 +72,21 @@ case class TextInput[F[_]: Functor] (path: Path, docType: TextDocumentType, inpu
 
 object TextInput {
 
-  private def readAll[F[_]: Sync](reader: Resource[F, Reader], sizeHint: Int): F[String] = 
-   reader.use(InputRuntime.readAll(_, sizeHint))
+  private def readAll[F[_]: Sync](input: fs2.Stream[F, Byte], codec: Codec): F[String] = 
+    input.through(fs2.text.decodeWithCharset(codec.charSet)).compile.string
```
armanbilge, line 76:
```suggestion
  private def readAll[F[_]: Concurrent](input: fs2.Stream[F, Byte], codec: Codec): F[String] = 
    input.through(fs2.text.decodeWithCharset(codec.charSet)).compile.string
```

FS2 has both a `Concurrent` compiler and a `Sync` compiler, but using the `Sync` compiler for a potentially `Concurrent` stream is not ideal (http4s has an internal lint to prevent this "downgrade"). See discussion in https://github.com/typelevel/fs2/pull/2657.

io/src/main/scala/laika/io/runtime/VersionedLinkTargets.scala
```diff
@@ -72,9 +72,10 @@ private[runtime] object VersionedLinkTargets {
   private def loadVersionInfo[F[_]: Sync] (existing: BinaryInput[F]): F[Map[String, Seq[Path]]] = {
     def asSync[T](res: ConfigResult[T]): F[T] = Sync[F].fromEither(res.leftMap(ConfigException.apply))
     
-    InputRuntime
-      .textStreamResource(existing.input, Codec.UTF8)
-      .use(InputRuntime.readAll(_, 8096))
+    existing.input
+      .through(fs2.text.utf8.decode)
+      .compile
```
armanbilge, line 77: `Sync` compiler being used here I think.

io/src/main/scala/laika/io/runtime/RendererRuntime.scala
```diff
@@ -128,6 +128,10 @@ object RendererRuntime {
       
       val styles =  finalRoot.styles(fileSuffix) ++ getThemeStyles(themeInputs.parsedResults)
       val pathTranslator = createPathTranslator(translatorConfig, Root / "dummy", lookup).translate(_:Path)
+
+      def createDirectory (file: File): F[Unit] =
+        Sync[F].delay(file.exists || file.mkdirs()).flatMap(if (_) Sync[F].unit
```
armanbilge, line 133:
```suggestion
        Sync[F].blocking(file.exists || file.mkdirs()).flatMap(if (_) Sync[F].unit
```

Why not use `Files[F]` here?

io/src/main/scala/laika/render/epub/ZipWriter.scala
```diff
@@ -37,19 +35,18 @@ object ZipWriter {
     * file (called `mimeType`) is written uncompressed. Hence this is not
     * a generic zip utility as the method name suggests.
     */
-  def zipEPUB[F[_]: Sync] (inputFs: Seq[BinaryInput[F]], output: BinaryOutput[F]): F[Unit] = {
+  def zipEPUB[F[_]: Sync] (inputs: Seq[BinaryInput[F]], output: BinaryOutput[F]): F[Unit] = {
```
armanbilge, line 38: I wonder if some of this `Zip` stuff could be upstreamed into FS2. I think it came up on the Discord the other day.

io/src/main/scala/laika/render/epub/ZipWriter.scala
```diff
@@ -37,19 +35,18 @@ object ZipWriter {
...
+    def copyAll (zipOut: ZipOutputStream, pipe: fs2.Pipe[F, Byte, Nothing]): F[Unit] = {
     
-      def writeEntry (input: (InputStream, Path), prepareEntry: ZipEntry => F[Unit] = _ => Sync[F].unit): F[Unit] = for {
-        entry <- Sync[F].delay(new ZipEntry(input._2.relative.toString))
+      def writeEntry (input: BinaryInput[F], prepareEntry: ZipEntry => F[Unit] = _ => Sync[F].unit): F[Unit] = for {
+        entry <- Sync[F].delay(new ZipEntry(input.path.relative.toString))
         _     <- prepareEntry(entry)
         _     <- Sync[F].blocking(zipOut.putNextEntry(entry))
-        _     <- CopyRuntime.copy(input._1, zipOut)
+        _     <- input.input.through(pipe).compile.drain
```
armanbilge, line 46: Sync compiler again I suspect.

io/src/main/scala/laika/render/epub/ZipWriter.scala
```diff
@@ -60,18 +57,14 @@ object ZipWriter {
       }
 
       writeEntry(inputs.head, prepareUncompressedEntry) >> 
-        inputs.tail.map(writeEntry(_)).sequence >> 
+        inputs.toList.tail.map(writeEntry(_)).sequence >>
```
armanbilge, line 60:
```suggestion
        inputs.toList.tail.traverse(writeEntry(_)) >> 
```

io/src/main/scala/laika/helium/generate/MergedCSSGenerator.scala
```diff
@@ -16,13 +16,19 @@
...
 import laika.ast.Path.Root
-import laika.io.model.InputTree
+import laika.io.model.{BinaryInput, InputTree}
 
 private[helium] object MergedCSSGenerator {
 
+  private def merge[F[_]: Sync](inputs: Seq[BinaryInput[F]]): F[String] =
+    inputs.toList
+      .map(_.input.through(fs2.text.utf8.decode).compile.string)
+      .sequence
```
armanbilge, line 29:
```suggestion
      .traverse(_.input.through(fs2.text.utf8.decode).compile.string)
```

Also note the sync compiler used here.

io/src/main/scala/laika/io/model/Input.scala
```diff
@@ -16,44 +16,45 @@
...
-    val resource = Resource.fromAutoCloseable(Sync[F].delay(new BufferedInputStream(new FileInputStream(file))))
-    BinaryInput[F](path, resource, targetFormats, Some(file))
+    val stream = Files[F].readAll(fs2.io.file.Path.fromNioPath(file.toPath))
+    BinaryInput[F](path, stream, targetFormats, Some(file))
   }
 
   def fromStream[F[_]: Sync] (path: Path, stream: F[InputStream], autoClose: Boolean, targetFormats: TargetFormats = TargetFormats.All): BinaryInput[F] = {
-    val resource = if (autoClose) Resource.fromAutoCloseable(stream) else Resource.eval(stream)
-    BinaryInput(path, resource, targetFormats)
+    val input = fs2.io.readInputStream(stream, 8096, autoClose)
```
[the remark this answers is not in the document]
jenshalm: True, I found traces of either 32K or 64K as defaults in the code base. `Files[F].readAll` uses 64K, so I could use that here simply for consistency. I would assume (hope) the fs2 devs arrived at these values via some benchmarking. 🙂

io/src/main/scala/laika/io/model/Input.scala
jenshalm: Oh, that is quite subtle and somewhat counterintuitive - something I wasn't aware of. I also could not find references to this in the fs2 docs.

io/src/main/scala/laika/io/runtime/RendererRuntime.scala
jenshalm: This part gets migrated in the next PR where creating and scanning directories is handled. Leaving this as it is, for a smoother merge.

io/src/main/scala/laika/render/epub/ZipWriter.scala
jenshalm: That would be an interesting API design challenge, given how much metadata you'd need to route through such a stream. Also note this funny requirement that EPUB has: the first entry needs to be uncompressed. The API would need to support these kind of edge cases.

io/src/main/scala/laika/io/model/Input.scala
armanbilge: Yeah you are absolutely right 😕 I guess it's sort of documented here:
https://github.com/typelevel/fs2/blob/9db8a220a07fc5b203eff4d311f53f769fb66a5d/core/shared/src/main/scala/fs2/Compiler.scala#L54

armanbilge: (But please open an issue about how we might improve that.)

jenshalm:
Thanks a lot, that was very helpful. In particular, I wasn't aware of those fs2 compiler subtleties. And apparently `traverse` is not as ingrained in my memory as in the "the answer is always traverse" camp. 😆 

I addressed all comments except for one which is deferred to the next PR.

io/src/main/scala/laika/helium/generate/MergedCSSGenerator.scala
```diff
@@ -16,13 +16,18 @@
...
-import cats.effect.Async
+import cats.effect.{Async, Sync}
 import cats.implicits._
 import laika.ast.Path.Root
-import laika.io.model.InputTree
+import laika.io.model.{BinaryInput, InputTree}
 
 private[helium] object MergedCSSGenerator {
 
+  private def merge[F[_]: Sync](inputs: Seq[BinaryInput[F]]): F[String] =
```
armanbilge, line 26:
```suggestion
  private def merge[F[_]: Concurrent](inputs: Seq[BinaryInput[F]]): F[String] =
```

io/src/test/scala/laika/io/helper/ResultExtractor.scala
```diff
@@ -33,14 +28,8 @@ trait ResultExtractor {
...
-    // TODO - move to InputRuntime
-    private def readText (stream: Resource[F, InputStream]): F[String] = {
-      stream.flatMap { str =>
-        InputRuntime.textStreamResource(Sync[F].pure(str), Codec.UTF8, autoClose = false)
-      }.use { reader =>
-        InputRuntime.readAll(reader, 4000)
-      }
-    }
+    private def readText (stream: fs2.Stream[F, Byte]): F[String] = 
+      stream.through(fs2.text.utf8.decode).compile.string
```
armanbilge, line 32: Another sync compiler, although this is test sources.

armanbilge (commented):
Glad to have been of help :) yeah the compiler stuff is unfortunately subtle :( I checked through again and I think there's one last one.

armanbilge (approved):

PR #286, opened 2022-05-05
Helium: automatically set canonical link for versioned pages via JavaScript

### LINKED ISSUE #264: Add `<link rel="canonical" href="...">` to versioned pages
We've been discussing this in https://github.com/http4s/http4s/issues/5715#issuecomment-1059806899.

The basic idea is that when there are multiple versions of the same page, a canonical url helps search engines index them as a single entry which is good for SEO. This technique is adopted by e.g. https://haha.readthedocs.io/en/latest/canonical.html

At first glance, this seems easy to support by adding an additional option to the `Versions` config where the canonical version can be specified and updating the template. Thinking a bit harder, it's not obvious what to do about pages that are present in the current version but not present in the canonical version.

jenshalm:
**Configuration and behaviour**

This PR adds a new property `canonical: Boolean` to the `Version` configuration class. 

If set to `true` the Helium theme's JavaScript will insert a canonical link into the `head` element of the page
(`<link rel="canonical" href="..."`) under the following conditions:

* The canonical version contains a page with the same local path (the path segments after the version segment).
* The page does not already contain a canonical link (this can happen if it was set manually via Laika's `laika.metadata.canonicalLink` attribute in the frontmatter of the markup, or by using a custom template that contains such a link)

Ideally users should always define the base URL of the site (e.g. via `.site.baseURL` in the Helium config API). If it is not set Laika will insert a relative link to the canonical page, which is legal, but not recommended practice.

**Validation**

This PR adds validation to the `Versions` configuration class, it ensures that:
* Only one configured version has the canonical flag set to true
* Every version declares a unique path segment for building URLs (this is unrelated to this PR, but was missed in earlier releases)

It did not seem feasible to let the type system help with detecting these problems at compile time, therefore it is a runtime check (transformation will fail if these conditions are not met).

**Reasons for not statically generating the link**

Laika's versioning configuration follows a recommended pattern of not requiring up-to-date and accurate version configuration in maintenance branches. Making this a requirement would be error-prone and a maintenance burden for site authors and most likely always lead to some branches being out of sync at least some of the time. For this reason the full version info is not known during site generation on a maintenance branch. When the page loads, Laika loads the `versionInfo.json` generated and published from the main branch and, at this point, has all the information needed to generate the link. Documentation for Google Search suggests that crawlers do execute JavaScript before indexing a page, therefore making this dynamic should not be a problem in practice.

Closes #264


armanbilge:
> If it is not set Laika will insert a relative link to the canonical page, which is legal, but not recommended practice.

Thanks, this is good to know. I see that it is covered here:
https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-link-method

armanbilge:
> The canonical version contains a page with the same local path (the path segments after the version segment).

This is interesting. Suppose a new version v0.42 or something is published as the canonical, without a page that was previously in v0.40 and v0.41. IIUC prior to v0.42 that page would have a canonical version. However, upon v0.42 becoming canonical, this page will no longer have a canonical. So the previously "linked" v0.40 and v0.41 versions of that page will become separate entities again. Which is not ideal.

jenshalm:
> Suppose a new version v0.42 or something is published as the canonical, without a page that was previously in v0.40 and v0.41. IIUC prior to v0.42 that page would have a canonical version. However, upon v0.42 becoming canonical, this page will no longer have a canonical. So the previously "linked" v0.40 and v0.41 versions of that page will become separate entities again. Which is not ideal.

I think I see this as expected behaviour, unless you want to implement a crystal ball in Laika. 🙂 Remember that the auto-linking is a convenience feature and users are expected to set the canonical version manually for all edge cases.

armanbilge:
Hm, not sure if I would go so far to call it expected behavior :)

> users are expected to set the canonical version manually for all edge cases.

I mean, sure, except that if in v0.42 I decide to remove a page that's been around since v0.1 I have to manually fix all of those. Better not remove the page to begin with!

... but on second thought, maybe this is a good thing. Never remove a page, but instead replace it with something that explains why there is no longer a page there.

> unless you want to implement a crystal ball in Laika

Huh, I _thought_ that's what it was 😛 



In all seriousness, some sort of canonical priority scheme would work, no? Check v0.42 first, then v0.41, etc. to try and identify a canonical for a page. But maybe not worth.

jenshalm:
> In all seriousness, some sort of canonical priority scheme would work, no? Check v0.42 first, then v0.41, etc. to try and identify a canonical for a page. But maybe not worth.

I don't feel like that would be a good idea.

First, it's complicated to implement and complicated to explain and document, for something that only covers some (probably rare) edge case.

Secondly, my suspicion is in the majority of cases it wouldn't even be what the user wants! Why would a page disappear? If you moved it, I'm sure you want to point to the new location and not to some stale version. Same if you merged two pages. Only if content really disappears, e.g. because a feature module had been phased out, this is what you'd want. But that's not the most likely scenario. So it's an edge case within an edge case scenario. 

That's why you'd need a true crystal ball to determine the user's intention, and some poor space dog's helmet doesn't cut it in this case. 🙂

armanbilge:
> Why would a page disappear? If you moved it, I'm sure you want to point to the new location and not to some stale version. Same if you merged two pages.

Right, thanks, these do seem like realistic scenarios and I understand better what you mean about the crystal ball :) but unless you go back and manually set all the canonicals to point to the new location, all the old versions of the page will become disconnected. But it is very difficult for Laika to do anything about that, without some additional complicated canonical forwarder config. From the user side, it still seems like the best thing is to _never_ remove a page, only replace it with a page that explains where the content is now.

jenshalm:
Yes, that sums it up well. The only way to further support content re-structuring would be to centrally define aliases. That could be used for two purposes: for the canonical link and for the version switcher, too (it goes to the configured fallback link for that version if the same path does not exist, otherwise you really just switch versions without leaving the page).

If a lot of our users ask for such an alias system, I'd consider it, but that would be 0.20 or 1.0.

PR #290, opened 2022-05-13
Deprecate API expecting java.io.File in favour of FilePath + introduce effectful FileFilter

jenshalm:
This PR is not strictly part of the fs2 migration, but related in the effort to reduce the use of Java APIs in Laika's public API. It's a collection of related changes listed below:

**Deprecate API expecting `java.io.File` in favour of Laika's own `FilePath` type**

* Methods in user-facing public APIs that expect as a parameter a single instance of type `java.io.File` are deprecated
* Methods in user-facing public APIs that expect a `Seq[File]` are modified without deprecation, due to type erasure preventing an overload of the deprecated and new method. Since APIs taking a single file are probably more commonly used, this breaking change is hopefully acceptable.
* Methods in public-internal APIs are simply changed without deprecation.

**Introduce a new `FileFilter` trait**

Previously some APIs expected an argument of type `File => Boolean` which was limited in that it did not allow for effectful filter logic. This PR introduces a new trait that overcomes that limitation and uses it in APIs where the pure function has previously been used:

```scala
trait FileFilter {
  def filter[F[_] : Async] (file: FilePath): F[Boolean]
}
```

**Other small changes**

* Deprecates `InputOps.fromWorkingDirectory` and `TextOutputOps.toWorkingDirectory` in favor of asking users to pass a relative `FilePath`.
* The order of arguments for the builder methods in `TextInput`, `TextOutput`, `BinaryInput` and `BinaryOutput` have been aligned with the order of similar methods in the user-facing APIs.

### DIFF
--- a/io/src/main/scala/laika/io/config/ResourceLoader.scala
+++ b/io/src/main/scala/laika/io/config/ResourceLoader.scala
@@ -16,15 +16,16 @@
 
 package laika.io.config
 
-import java.io.{File, FileNotFoundException, InputStream}
-import java.net.URL
 import cats.effect.{Async, Sync}
 import cats.syntax.all._
+import fs2.io.file.Files
 import laika.ast.DocumentType
 import laika.ast.Path.Root
 import laika.config.ConfigResourceError
-import laika.io.model.TextInput
+import laika.io.model.{FilePath, TextInput}
 
+import java.io.{FileNotFoundException, InputStream}
+import java.net.URL
 import scala.io.Codec
 
 /** Internal utility for loading text resources from the file system, the classpath or via its URL.
@@ -40,7 +41,7 @@ object ResourceLoader {
     * successfully parsed resources will be returned as `Some(Right(...))`.
     */
   def loadFile[F[_]: Async] (file: String): F[Option[Either[ConfigResourceError, String]]] = 
-    loadFile(new File(file))
+    loadFile(FilePath.parse(file))
 
   /** Load the specified file (which may be a file on the file system or a classpath resource).
     *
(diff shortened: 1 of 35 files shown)


armanbilge:
Do you mind if I ask a design question? I often wonder why Laika APIs choose this pattern:

```scala
trait FileFilter {
  def filter[F[_] : Async] (file: FilePath): F[Boolean]
}
```

versus this one:
```scala
trait FileFilter[F[_]] {
  def filter(file: FilePath): F[Boolean]
}
```

Thanks :)

jenshalm:
That's a good question. 🙂 Unfortunately I don't fully remember why I started down that path.

First, it's not consistently the other pattern, for many types the library instantiates internally it's the other way round (e.g. `TextInput[F]`, `TreeParser[F]`). But when it comes to types passed to the library by the user I actually had issues with using the other pattern. I think it was `ThemeProvider` where I first had the other pattern, but was struggling to please the compiler in some scenarios where it claimed not being able to prove that one `F` corresponds to some other `F`. I would need to play with it to remember it. There were probably ways around it, but since I did not see any significant downside in using the other pattern, I took a shortcut instead.

armanbilge:
Thanks, I see, that's very odd 🤔 

> but since I did not see any significant downside in using the other pattern

Often it's recommended to avoid proliferation of powerful constraints such as `Sync` and `Async`. Typically moving the effect type parameter to the trait helps to keep them encapsulated. However this is not the right pattern for every library :)

jenshalm:
I see, although I'm afraid the nature of `laika-io` generally does not leave that much room for weaker constraints (moving to fs2 actually shifted it further to stronger ones).

I'd love to remember what the issues were, too. When I have some time, I might do a demo branch where I revert `ThemeProvider` to see what the issue was.

armanbilge:
> I see, although I'm afraid the nature of `laika-io` generally does not leave that much room for weaker constraints (moving to fs2 actually shifted it further to stronger ones).

Note that `fs2.io.file.Files[F]` is indeed a weaker constraint than `Async[F]` :) ideally most of `laika-io` can be written relying only on `Concurrent` and `Files`.

As a case study, consider:
https://github.com/planet42/Laika/blob/0a70c44e81e504940ae30a1b843ccf4168e9f9e6/io/src/main/scala/laika/io/config/ResourceLoader.scala#L34

By turning this into a `trait ResourceLoader[F[_]]`, then whenever a user needs its capabilities instead of adding an `Async` constraint (give me the power do to anything!) they can simply add a `ResourceLoader` constraint (give me the power to load resources!). Of course, to instantiate a `ResourceLoader` requires `Async` (as does instantiating `Files`), but this doesn't need to happen until the very top of the application.

jenshalm:
I think what you describe applies more broadly to the use cases of libraries like http4s, fs2 or cats-effect, where what you describe adds significantly more value for users coding against those APIs. The scenario for Laika is a bit different. If you obtain a type from the laika-io module, you are probably not going to pass it down through your beautifully designed tagless-final application layers, but merely use it to describe a customization and then directly pass it back to the Laika API. And the place where you do that is most commonly the transformer API where you have to pass an Async anyway (at that entry point there is really no way around it). And those entry-point APIs need to remain accessible for people who barely know what an `IO` is and only copy a code sample from the documentation.

We can revisit that in the 1.0 milestone series (which will most likely start either after 0.19 or 0.20, depending on user feedback). The first milestone will kick it off with a significant reduction of the public API surface. And for the types which survive that round, it might make sense then to see if their requirements can be reduced. The `ResourceLoader` you mentioned won't be one of them. 🙂 It's public "by accident" really.

PR #293, opened 2022-05-19
add shell syntax highlight

i10416:


core/shared/src/main/scala/laika/parse/code/languages/ShellSyntax.scala
```diff
@@ -0,0 +1,88 @@
...
+    Keywords(BooleanLiteral)("true", "false"),
+    Keywords(
+      "break","case","continue","do","done","echo","else","esac","eval",
+      "exec","exit","export","fi","for","function","if","read","readonly","return",
+      "set","select","shift","then","trap","unlimit","unmask","unset","until","wait","while","!",
+    ),
+    Identifier.alphaNum,
+    NumberLiteral.decimalInt,
+    NumberLiteral.decimalFloat,
+    StringLiteral.Substitution.between("$(", ")"),
```
jenshalm, line 76: It might be good to support arithmetic expressions explicitly. That would be `between("$((", "))")`. The expression above does not cover that as it would end on the first closing parenthesis, not the second. It would need to be added before the above, as it would otherwise be shadowed.

jenshalm (changes requested):
Thanks for adding another syntax highlighter. Reviews of those always take a bit longer as I cannot just quickly review the code itself (which was an easy task in this case), but also need to study relevant reference documentation for some basic compliance checks.

This looks fairly close to being good to merge. There are just a few suggested changes in the comments below. It would also be good to immediately add the new syntax to the two places where the manual lists supported languages (like you recently did for Dart).

In terms of timing, the 0.19 release is still about 6-8 weeks away, so there is still plenty of time to land this.

jenshalm:
Superseded by #422 which is the original commit from this PR plus all requested changes.

PR #294, opened 2022-05-20
Update build.sbt

i10416:
sync to latest version

Laika v0.18.2 has already published.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import sbt.Keys.{artifactPath, crossScalaVersions}
 import Dependencies._
 
 lazy val basicSettings = Seq(
-  version               := "0.18.1",
+  version               := "0.18.2",
   homepage              := Some(new URL("https://planet42.github.io/Laika/")),
   organization          := "org.planet42",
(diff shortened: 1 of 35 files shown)


jenshalm:
Apologies about the confusion, but this is unfortunately not correct. 0.18.2 had been released from a branch (0.18.x). That branch already contains most of the changes you made here, and if we ever push updates for the docs for 0.18.x it will be from that branch.

The only thing we should do on the main branch is to update the version in the build to 0.19.0-SNAPSHOT, which I forgot to do after creating the 0.18.x branch. If you use a released version here, you might accidentally break the test suite for the sbt plugin which then might pull in released artifacts from the build cache instead of building it fresh from the local sources.

But for setting it to 0.19.0-SNAPSHOT the sbt plugin test suite has to be adjusted first for some of the recent changes. If you want to do that, feel free to do that in a new PR (we cannot keep any of the changes from this one), but it's not pleasant (it's often just tracking down weird white-space differences introduces by template changes), so if you don't want to do that you can also close this and let me look into it in the coming days.

PR #295, opened 2022-05-20
mention Dart syntax highlight for next release

i10416:


jenshalm:
Thank you, I somehow thought I did this already, but apparently I didn't.

Note that the supported syntax is listed in two places, would be nice to always add it to both.
The other is here: https://github.com/planet42/Laika/blob/master/docs/src/01-about-laika/01-features.md#content-organization

i10416:
> Note that the supported syntax is listed in two places,

Thanks, done.

PR #297, opened 2022-05-21
Skip duplicated titledoc segment

### LINKED ISSUE #296: breadcrumb behavior
Breadcrumb directive for shallow document tree generates duplicated navigation items. Is this intended behavior?

For example, when I have following tree and template,

```sh
.
└── dir
    └── README.md
```

```md

@:breadcrumb
```

```scala
val shallow = {
      val template = parseTemplate("@:breadcrumb").getOrElse(throw new Exception("Never"))
      DocumentTree(
        path = Root,
        content = Seq(
          DocumentTree(
            path = Root / "dir",
            content = Nil,
            titleDocument = Some(
              Document(
                path = Root/"dir"/"README",
                content = RootElement(Seq(Header(1,Seq(Text("Title"))))),
              )
            ),
          )
        ),
        templates = Seq(
          TemplateDocument(DefaultTemplatePath.forHTML,template)
        )
     )
}
```

I got

```
. . . NavigationList(Styles(breadcrumb)) - Elements: 3
. . . . NavigationItem(None,All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - '/'
. . . . . Content - Elements: 0
. . . . NavigationItem(Some(NavigationLink(ResolvedInternalTarget(/dir/README,#,All),true,All)),All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - 'Title'
. . . . . Content - Elements: 0
. . . . NavigationItem(Some(NavigationLink(ResolvedInternalTarget(/dir/README,#,All),true,All)),All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - 'Title'
. . . . . Content - Elements: 0
```

full tree
```
DocumentTreeRoot(DocumentTree(/,List(DocumentTree(/dir,List(),Some(Document(/dir/README,
RootElement - Blocks: 1
. TemplateRoot - TemplateSpans: 3
. . TemplateString - '|      |'
. . TemplateElement(0)
. . . NavigationList(Styles(breadcrumb)) - Elements: 3
. . . . NavigationItem(None,All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - '/'
. . . . . Content - Elements: 0
. . . . NavigationItem(Some(NavigationLink(ResolvedInternalTarget(/dir/README,#,All),true,All)),All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - 'Title'
. . . . . Content - Elements: 0
. . . . NavigationItem(Some(NavigationLink(ResolvedInternalTarget(/dir/README,#,All),true,All)),All,Styles(level1,breadcrumb))
. . . . . SpanSequence - Spans: 1
. . . . . . Text - 'Title'
. . . . . Content - Elements: 0
. . TemplateString - '|      |'
,Map(),laika.config.ObjectConfig@51772459,TreePosition(1))),List(),laika.config.EmptyConfig$@6ed6ca53,TreePosition(1))),None,List(),laika.config.EmptyConfig$@6ed6ca53,TreePosition.root),None,Map(),List(),Map())
```

i10416:
close #296


core/shared/src/main/scala/laika/directive/std/BreadcrumbDirectives.scala
```diff
@@ -56,19 +57,26 @@ object BreadcrumbDirectives {
...
         tree.parent match {
           case None => item :: items
           case Some(parent) => entriesFor(parent,item ::items)
         }
       }
 
-      val docEntry = {
+      // if cursor is targeting title document, we don't need to append docEntry
+      // as Seq[NavigationItem] generated by `entriesFor` contains link to title document
+      if(cursor.target.path.name.eqv(TitleDocumentConfig.defaultInputName)) {
```
i10416, line 71: Ah, I've noticed I should use title document name from config as user may choose another name but README🤔

jenshalm: I think the easiest way to check this is `if (cursor.parent.titleDocument.exists(_.path == cursor.path))`

i10416: That seems much simpler. Thanks!

core/shared/src/test/scala/laika/directive/std/NavigationDirectiveSpec.scala
```diff
@@ -408,7 +414,287 @@ class NavigationDirectiveSpec extends FunSuite with ParagraphCompanionShortcuts
     runDocument(input,docList(Root / "tree-1" / "doc-3", 3, 1))
   }
 
+  test("template breadcrumb directive - breadcrumb skips TitleDocument docEntry"){
```
jenshalm, line 417: It would be good to only add this one test case to the spec. The tests are very verbose and different in style than the existing tests and most of them below cover scenarios already covered by the existing tests, except for the last one which is a scenario that does not warrant an extra test.

If we reduce the size of the spec and switch to the simpler title document condition check as discussed above, this would ready to be merged.

jenshalm (changes requested):

core/shared/src/test/scala/laika/directive/std/NavigationDirectiveSpec.scala
i10416: Thanks!

jenshalm (approved):

PR #312, opened 2022-08-20
update dependencies

jenshalm:


armanbilge:
If you're interested, you can sign up for scala-steward here :)

https://github.com/VirtusLab/scala-steward-repos

jenshalm:
Yes, I know 🙂 - I think I need some time to pass from my last daytime gig where it was more noise and hassle than help.

PR #322, opened 2022-08-27
preview server - reduce graceful shutdown to 2 seconds

jenshalm:
Also closes #311 as the upgrade of cats-effect made the observed issue with exceptions when navigating go away.
I'll move to automatic updates once 0.19 is out.

### DIFF
--- a/sbt/src/main/scala/laika/sbt/Tasks.scala
+++ b/sbt/src/main/scala/laika/sbt/Tasks.scala
@@ -232,7 +232,7 @@ object Tasks {
       System.in.read
     }
     finally {
-      streams.value.log.info(s"Shutting down preview server.")
+      streams.value.log.info(s"Shutting down preview server...")
       cancel.unsafeRunSync()
     }
   }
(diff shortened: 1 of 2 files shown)


sbt/src/main/scala/laika/sbt/Tasks.scala
```diff
@@ -232,7 +232,7 @@ object Tasks {
       System.in.read
     }
     finally {
-      streams.value.log.info(s"Shutting down preview server.")
+      streams.value.log.info(s"Shutting down preview server...")
```
armanbilge, line 235: These dots are ominous 😂

armanbilge (approved):
Thanks for reconsidering this, much appreciated :) and please let us know if you encounter any other issues, we've been rolling out several improvements/fixes lately.

sbt/src/main/scala/laika/sbt/Tasks.scala
jenshalm: Just supporting the most impatient souls amongst us, so that they don't think something is broken while the 2 seconds pass... 🙂

PR #325, opened 2022-08-29
Helium: smarter defaults for home link + better error message

jenshalm:
The enhanced mechanism will attempt to set the home link in the following order:

* A target that has been set explicitly via the existing option in the Helium API (`.site.topNavigationBar(homeLink = ...)`)
* The landing page of the site if available (configured via `Helium.defaults.site.landingPage(...)`)
* Any `README.md` or `README.rst` files in the root directory of the input sources.
  The page recognized as the title page per directory might be different in case the user overrides
  the `laika.titleDocuments.inputName` setting.
* If the home link is neither configured explicitly nor any of the default home link targets mentioned above exist
  in the input tree, the transformation will fail with an error pointing to this part of the documentation instead of the existing, unhelpful error message.

@noelwelsh, @Ostrzyciel - this should improve the situation you reported in #309 - there will be fewer scenarios where the error will occur, and if it does there will be a clearer message.

### DIFF
--- a/docs/src/03-preparing-content/03-theme-settings.md
+++ b/docs/src/03-preparing-content/03-theme-settings.md
@@ -530,20 +530,33 @@ Helium.defaults.site
 ```
 
 All the properties shown above have default values, so you don't have to specify them all.
-The link to the homepage can be customized, by default it is pointing to `index.html` and using the Helium home icon.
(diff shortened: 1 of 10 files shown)


noelwelsh:
👍

PR #352, opened 2023-01-03
mima-setup for 0.19.x maintenance

jenshalm:
The 1.0 milestone series will switch to CI releases and will see a wider range of build changes, but for 0.19 maintenance I intend to keep the current manual approach and only add a minimal mima setup (scalafmt and mdoc will follow before branching).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           
       - name: Build and test
         run: |
-          sbt -v +test mimaReportBinaryIssues scripted
+          sbt -v +test +mimaReportBinaryIssues scripted
           rm -rf "$HOME/.ivy2/local" || true
           find $HOME/.ivy2/cache                       -name "ivydata-*.properties" -delete || true
           find $HOME/.ivy2/cache                       -name "*-LM-SNAPSHOT*"       -delete || true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           
       - name: Build and test
         run: |
-          sbt -v +test scripted
+          sbt -v +test mimaReportBinaryIssues scripted
           rm -rf "$HOME/.ivy2/local" || true
           find $HOME/.ivy2/cache                       -name "ivydata-*.properties" -delete || true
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -30,7 +30,8 @@ def priorTo2_13(version: String): Boolean =
   }
 
 lazy val moduleSettings = basicSettings ++ Seq(
-  crossScalaVersions := Seq(versions.scala2_12, versions.scala2_13, versions.scala3)
+  crossScalaVersions    := Seq(versions.scala2_12, versions.scala2_13, versions.scala3),
+  mimaPreviousArtifacts := Set(organization.value %% moduleName.value % "0.19.0")
```
armanbilge, line 34: I think you may want `%%%` here, otherwise you end up comparing JS artifacts to JVM artifacts 🤔 

I do something a bit different here, which I think is more robust (works with all artifacts, including sbt plugins).
https://github.com/typelevel/sbt-typelevel/blob/9c6b9b19080d02a25c94e03ae6535a4c30e8b709/mima/src/main/scala/org/typelevel/sbt/TypelevelMimaPlugin.scala#L88

Btw really great to see this!

jenshalm: Excellent, I kind of felt that there must be an easier way! 🙂 

And you are right, when I re-ran after the changes it actually downloaded the Scala.js artifacts which it apparently did not do before...

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
           
       - name: Build and test
         run: |
-          sbt -v +test scripted
+          sbt -v +test mimaReportBinaryIssues scripted
```
armanbilge, line 32: Assuming you want to check for all Scala versions.

```suggestion
          sbt -v +test +mimaReportBinaryIssues scripted
```

jenshalm: Yes, makes sense. But I do wonder in which scenario a check for one version could fail, but not for the other, when there are no version-specific sources?

armanbilge: Lots of obscure scenarios :) if all scala versions emitted identical binary APIs, then we wouldn't need to do all this cross-building in the first place! There are differences in the encodings that sometimes mean a binary-compatible change in one version is not a binary-compatible change in another version.

Issue #360, opened 2023-01-18
Develop a Laika CLI for Node.js and Native


armanbilge:
This issue is spun out of https://github.com/planet42/Laika/issues/281#issue-1220788257.

> - **Postpone support for Scala.js in laika-io**
> 
>   This option has been mentioned in the original request. It is, however, beyond the scope of the 0.19 roadmap. Any future consideration would require that someone opens a ticket for this and is able to generate significant interest in the community for having this module on Scala.js.

The goals of this issue are to propose the broader idea, evaluate if it is even within Laika's scope, and see if there is community interest.

### Proposal

The idea is really two separate but (IMO) related things.
- cross-publish laika-io for Scala.js (targeting Node.js) and Scala Native
- create a CLI application

The motivation for a CLI is so that Laika can be used as a standalone site generator à la Jekyll. @valencik and I recently discussed this in context of the typelevel.org website. I'm not entirely sure if this is possible, since I've really only used it via sbt: specifically, if there is enough configuration flexibility via HOCON/templates, without resorting to Scala code.

Assuming a CLI is a feasible and attractive feature, then cross-building laika-io to Scala.js and Native becomes very interesting.
- Scala Native CLI would be a self-contained lightweight binary with instant startup (vs a JVM application)
- A Scala.js CLI published to NPM would be extremely easy to install and run from virtually everywhere

If publishing a CLI application is out-of-scope for Laika, then I think cross-building laika-io for JS/Native is less valuable. The core is already cross-published for JS (and maybe Native someday) which, if desired, can be used directly with fs2-io without going through laika-io.

jenshalm:
Many thanks for authoring this proposal. While I generally think this presents an interesting and exciting opportunity, I also have quite a number of doubts, and they are less about the feasibility of this effort (I really cannot judge feasibility of anything related to Scala Native), but more about a) the scope of work that I would see as mandatory to make this an attractive offering for which I'll try to provide some context below and b) the fact that a CLI application is not a frequently requested feature at the moment.

Since this comment turned out to become rather long, I've split it into three main section that address A) the target audience you intend to reach with the proposal, B) the scope of work I'd see as necessary to produce an attractive offering and C) the question of development and maintenance, specifically in the context of the other work happening in the coming months.


# Target Audience

[...]

PR #368, opened 2023-02-10
Attempt at adding Native support for core

zetashift:
I have no idea what I'm doing :P.

I basically copy pasted the JVM implementation 1:1, and trying to see how far I get from there.
If anything at all can be improved let me know

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -104,7 +104,7 @@ lazy val docs = project.in(file("docs"))
     Laika / target := baseDirectory.value / "target"
   )
 
-lazy val core = crossProject(JSPlatform, JVMPlatform)
+lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .withoutSuffixFor(JVMPlatform)
   .crossType(CrossType.Full)
   .in(file("core"))
--- a/core/native/src/main/scala/laika/time/PlatformDateTimeImpl.scala
+++ b/core/native/src/main/scala/laika/time/PlatformDateTimeImpl.scala
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+
+import cats.syntax.all._
+import java.time.format.{DateTimeFormatter, DateTimeFormatterBuilder, FormatStyle}
+import java.time.{LocalDateTime, OffsetDateTime, ZoneId}
+import java.util.Locale
+import scala.util.Try
+
+/**
+  * @author Jens Halm
+  */
+object PlatformDateTimeImpl extends PlatformDateTime {
+
+  type Type = OffsetDateTime
+
+  private[laika] def now: Type = OffsetDateTime.now()
+
+  private val offsetDateTime: DateTimeFormatter = new DateTimeFormatterBuilder()
+    .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
+    .appendPattern("[XXX][X]")
+    .toFormatter
+
+  def parse (dateString: String): Either[String, Type] = {
+
+    def parseOffsetDateTime(input: String): Either[String, Type] = Try {
+      OffsetDateTime.from(offsetDateTime.parse(input))
+    }.toEither.left.map(_.getMessage)
+
+    def parseLocalDateTime(input: String): Either[String, Type] = Try {
+      OffsetDateTime.from(LocalDateTime.parse(input).atZone(ZoneId.systemDefault()))
+    }.toEither.left.map(_.getMessage)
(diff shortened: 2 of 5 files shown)


armanbilge:
> I basically copy pasted the JVM implementation 1:1, and trying to see how far I get from there.

Instead of copying, you can place these sources in the `jvm-native` folder so that they can be shared :)

jenshalm:
Welcome, and many thanks for looking into this. However, before you spend any more time on this I'd like to point you to the fairly recent debate in #360, which will make it clear that this is unlikely to get merged in the near future.

The short summary is that we established that increasing the platform support for Laika is valuable only if we also look into developing a CLI application, but the conditions for that are currently not met (not enough interest in the community, too few maintainers for Laika, nobody committed to actually picking up the work for the CLI, which is much more effort that expanding Laika's platform support).

As things stand today, if I would make any change in Laika's platform support, I'd probably think about **removing** support for Scala.js. There are less than 10(!!) downloads per month, and the support does not only increase the build and release times, there were also several issues I had to deal with specifically to support this platform, which is not a reasonable way to spend my time when there are basically no users.

Nevertheless, I'm curious to know what your personal use case is for Laika on Native?

armanbilge:
> The short summary is that we established that increasing the platform support for Laika is valuable only if we also look into developing a CLI application

Oh, hm, that's not my conclusion :) I believe laika-core on Native would be valuable. But laika-io only makes sense with a CLI application. Quoting myself:

> If publishing a CLI application is out-of-scope for Laika, then I think cross-building laika-io for JS/Native is less valuable. The core is already cross-published for JS (and maybe Native someday) which, if desired, can be used directly with fs2-io without going through laika-io.

---

> As things stand today, if I would make any change in Laika's platform support, I'd probably think about removing support for Scala.js. There are less than 10(!!) downloads per month, and the support does not only increase the build and release times, there were also several issues I had to deal with specifically to support this platform, which is not a reasonable way to spend my time when there are basically no users.

I'm sorry to hear that.

jenshalm:
> > The short summary is that we established that increasing the platform support for Laika is valuable only if we also look into developing a CLI application
> 
> Oh, hm, that's not my conclusion :) I believe laika-core on Native would be valuable. But laika-io only makes sense with a CLI application. Quoting myself:
> 
> > If publishing a CLI application is out-of-scope for Laika, then I think cross-building laika-io for JS/Native is less valuable. The core is already cross-published for JS (and maybe Native someday) which, if desired, can be used directly with fs2-io without going through laika-io.

Sorry, I misunderstood that in a way that to me this logically expands to `laika-core` as it seems that at least for Scala.js there is no community use case for just supporting this module without expanding it to a CLI application. Where do you see a use case for just having `laika-core` on Native?

But in any case, my perspective on this remains: I am not considering adding maintenance for Native support to my todo list at this point, so until someone steps up and is willing to do the review and future maintenance for this, this cannot be merged.

armanbilge:
> Where do you see a use case for just having `laika-core` on Native?

The same use-cases as on JVM (minus the sbt plugin) e.g. server-side rendering of markdown. Unless the big picture here is that Laika itself doesn't really have a use-case outside of an sbt plugin?

---

> so until someone steps up and is willing to do the review and future maintenance for this, this cannot be merged

Right, that would be me :)

zetashift:
> However, before you spend any more time on this I'd like to point you to the fairly recent debate in https://github.com/planet42/Laika/issues/360, which will make it clear that this is unlikely to get merged in the near future.

Heyho! I indeed read that conversation and I completely understand the reasoning you give that a native target isn't easy in several ways and don't want to increase maintenance burden like that. I'm hoping that at some point in the future enough desire is garnered for a Scala Native target.

I put up this (draft) PR because I really like Scala Native and would like to increase its library support. And more specifically Laika, because I'm looking for a static site generator(or atleast markdown parser) with the attributes of being a tiny native executable and written in Scala.

Even if this never gets merged, it is still a great learning exercise for me and I must say the Laika code is very readable!

jenshalm:
> > Where do you see a use case for just having `laika-core` on Native?
> 
> The same use-cases as on JVM (minus the sbt plugin) e.g. server-side rendering of markdown. Unless the big picture here is that Laika itself doesn't really have a use-case outside of an sbt plugin?

Yes, sadly that is my impression at this point, judging by download numbers. To drop some of the numbers again (for November):

* sbt plugin - 11,000
* laika-core JVM 2.13 - 100
* laika-core JS 2.13 - 10

That's <1% and <0.1% of downloads respectively.

> > so until someone steps up and is willing to do the review and future maintenance for this, this cannot be merged
> 
> Right, that would be me :)

Ok, that would be an option then. But I'd still like to question whether it is worth it at this point in time and if you really want to add this to the apparently thousands of things you are involved in 🙂 .

jenshalm:
> Considering the above, it's a good question :) if you intend to keep Scala.js since it's already here, then adding Scala Native is probably straightforward. Of course, we have to see how this PR looks before making a final call on that.
> 
> Of course if you are seriously considering dropping Scala.js support altogether, then this doesn't make sense.

Ok, I'll leave the final call on this mostly to you then. Removing Scala.js support is not very likely at this point. While I do think that the past effort was not really worth it, a trigger for removing it would only be a significant number of future issues that are specific to JS while download numbers are still negligible. Only then I'd seriously think about it (and also probably propose it quite a long time before actually doing it).

Only requirements from my side at this point: If the final PR is binary compatible it can theoretically get merged anytime (meaning we do the shuffle of source folders for the 0.19 branch, too). If it can only be done for 1.0, as discussed, I'd prefer to defer the merging to a later phase of the milestone series.

Also @zetashift my next PR will introduce scalafmt which will also reformat the entire code base. It should be out in the coming 10 days or so, you may want to wait for that to avoid tons of merge conflicts.

PR #377, opened 2023-02-27
add scalafmt to build and reformat all sources

jenshalm:
This took a while as I was quite unhappy with some of the defaults and ended up reading the the documentation for every single configuration option.

Many settings are not worth discussing in detail, but there are two which will allow for a tiny amount of inconsistencies in formatting to prevent some of the outcomes I found undesirable:

### Scaladoc

I did not find an option that would wrap all lines in documentation comments that do exceed `maxColumn`, but leave everything else untouched. Instead the wrapping option would reflow the entire comment, which breaks the intended formatting style for scaladoc in this project: adding line breaks after periods and commas even if there is still space on the line to avoid a complete reflow in case of amendments. Without this the comments are harder to read and very often lead to larger diffs in PRs as adding a single word to the beginning of a comment can trigger a reflow of all subsequent lines. For this reason the option `docstrings.wrap = no` has been added.

### Select Chains

The default for scalafmt is that if a select chain does not fit onto a single line, then each select will be put onto a separate line. This is fine for most cases, but ruins the intended formatting for Helium configuration which is used a lot throughout the unit tests in the laika-io module. The Helium API always semantically groups two selects and I'd prefer to preserve this.

Example:

```scala
val theme = Helium.defaults
  .all.metadata(
    title = Some("Project Name"),
    language = Some("de"),
  )
  .epub.navigationDepth(4)
  .pdf.navigationDepth(4)
  .build
```

For this purpose the option `newlines.selectChains = keep` has been added.

These two options should be mentioned explicitly in the contributor's guide (which is also still on my todo list).


armanbilge:
In case you are not already aware of it—I recommend adding a `.git-blame-ignore-revs` with the formatting commit. 

Example:
https://github.com/typelevel/cats/blob/555a4f19a6d18f640223103d73ba15a5d74f2e58/.git-blame-ignore-revs

Of course, to be effective you will have to avoid the squash-merge for this PR :)

jenshalm:
Ah, nice, didn't know this. Will add this.

But first I need to figure out why `scalafmtCheckAll` fails on CI, but not locally...

PR #378, opened 2023-03-03
Change the Helium default for code font-size to be relative (fixing scenarios like code fonts in headlines)

TonioGela:
This is the upstreamed version of https://github.com/typelevel/sbt-typelevel/pull/472

It should address the fact that once you define a title that contains a `` block the font-size defaults to the code blocks' one, like [here](https://typelevel.org/cats/datatypes/validated.html#of-flatmaps-and-eithers).

I'm not sure which is the best css file in whom this change should be done, but as far as I understood `contents.css` contains most of the default style's settings.

Also, I got a red test testing it locally but I think it might have been a flaky timestamp related test

### DIFF
--- a/io/src/main/resources/laika/helium/css/content.css
+++ b/io/src/main/resources/laika/helium/css/content.css
@@ -67,14 +67,6 @@ h4, h5, h6 {
   margin-bottom: calc(var(--block-spacing) / 2);
 }
 
-/* This fixes the font-size of code snippets in headers */
-h1 > code { font-size: var(--header1-font-size); }
-h2 > code { font-size: var(--header2-font-size); }
-h3 > code { font-size: var(--header3-font-size); }
-h4 > code { font-size: var(--header4-font-size); }
-h5 > code { font-size: var(--header5-font-size); }
-h6 > code { font-size: var(--header6-font-size); }
-
 /* links =========================================== */
 
 a {
--- a/io/src/main/resources/laika/helium/css/content.css
+++ b/io/src/main/resources/laika/helium/css/content.css
@@ -67,6 +67,14 @@ h4, h5, h6 {
   margin-bottom: calc(var(--block-spacing) / 2);
 }
 
+/* This fixes the font-size of code snippets in headers */
+h1 > code { font-size: var(--header1-font-size); }
+h2 > code { font-size: var(--header2-font-size); }
+h3 > code { font-size: var(--header3-font-size); }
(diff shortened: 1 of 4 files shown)


TonioGela:
Someone made me noticed that pdf exports weren't fixed so I added the same code to `contents.epub.css` since I supposed that's the stylesheet used for epubs and pdfs.

jenshalm:
Hello and thanks for looking into this. I wasn't actually aware of the issue (or I forgot about it).

However, I think my preferred approach would be a more flexible one based on relative sizes. The approach is different for each output format, so I'll briefly go over them below:

**HTML**

Here we could simply switch the default to `0.9em`. This has two advantages: It solves the issue for all types of parent nodes, not just headlines and it looks a bit better (I find code fonts usually look nicer when they are a tad smaller than the regular font). This would be a simple one line change here: https://github.com/planet42/Laika/blob/main/io/src/main/scala/laika/helium/config/HeliumDefaults.scala#L125.

This would most likely require adjusting the expectations in a few Helium tests and a quick sanity check it looks ok in headlines, body text and code blocks.

**EPUB**

Here all font sizes are already expressed in relative values (https://github.com/planet42/Laika/blob/main/io/src/main/scala/laika/helium/config/HeliumDefaults.scala#L125).
This ensures that all font sizes adjust based on size the user has chosen in the reader. I didn't check it right now, but I'd expect that this issue does not exist for EPUB output.

**PDF**

This is sadly a bit complicated, as PDF is not styled based on web CSS (it has XSL-FO as the interim format, not HTML). I'm afraid I'd need to investigate a bit how to best solve it for PDF, as FO does not allow to set a default font size globally. You can ignore this in this PR if you want and only solve it for HTML for now.

TonioGela:
Sorry for the ultra-slow reply Jens. I fixed the implementation of HTML as you suggested and adapted the tests.
For what concerns PDF, I have to study further how XSL-FO works, so I'll propose to take just HTML and epub in consideration for now.

jenshalm (approved):

jenshalm:
Thanks for the update, the timing is perfectly fine as there are still a few weeks left to the 0.19.1 release.

This looks good to be merged now, we can defer the PDF solution to a separate PR, I still don't have a good idea.

jenshalm:
Just adjusted the title as it did no longer match the implementation. I will merge this within a few days unless someone finds any concerns with it.

Issue #379, opened 2023-03-04
Consider switching to a font without ligatures for code blocks


armanbilge:
Or, at least make it easier to configure this :) The problem is that ligatures obfuscate the actual characters you need to type.

j-mie6:
In particular, `Fira Mono` is a good ligature-less alternative for `Fira Code`

jenshalm:
I'm fine with switching to Fira Mono as the default, feel free to submit a PR 🙂 

Apart from that, in which way do you think the configuration should be easier?

armanbilge:
> Apart from that, in which way do you think the configuration should be easier?

https://github.com/planet42/Laika/blob/fad27dc8b90cea1a4b8128575ba76aea2d1e0d84/io/src/main/scala/laika/helium/config/settings.scala#L78-L80

I know there are excellent reasons for how this is working, it just means if you want to add a font you have to start from scratch :)

j-mie6:
> I'm fine with switching to Fira Mono as the default, feel free to submit a PR 🙂

Sure, I can do that. Do you prefer I remove the existing `Fira Code` resources or keep them in for ease of someone switching to the ligatures version just by using the `jar` resources (obviously, not keeping them in the embedding!). Note that the Fira Code `.ttf` files you have are not actually ones with ligatures (if you think it's a good idea to keep them around, I can try and find the right resource files for them too)

jenshalm:
I don't have a strong opinion either way, I am just a bit confused now. So there is a Fira Code without ligatures, that still differs from Fira Mono? If yes, what would be the reason for preferring one over the other?

j-mie6:
Well, I'm unsure, it's not clear _how_ you got a version without ligatures, because it's meant to have them! But at least, the ligatures don't render on PDFs generated by Laika (either because the OTF file is missing them, or the generation doesn't use them). Either way, the CSS used for the site definitely has them, and I'm not sure how to find one _without_ them.

armanbilge:
> So there is a Fira Code without ligatures, that still differs from Fira Mono? If yes, what would be the reason for preferring one over the other?

Seems it has some other nice-ities.

> Fira Code is not only about ligatures. Some fine-tuning is done for punctuation and frequent letter pairs.

<details>

![](https://github.com/tonsky/FiraCode/raw/master/extras/typographics.png)

</details>

j-mie6:
Ok so I checked, it just doesn't render ligatures in the PDFs for some reason: very very weird!

[...]

PR #380, opened 2023-03-04
`Fira Code` -> `Fira Mono`

j-mie6:
Switched fira-code over to fira-mono (which does not have ligatures), but retained the fonts just in case people want to switch back easily. As discussed in #379 

(also expanded the weights available for Fira Code in the resources whilst I'm at it


j-mie6:
Not sure what that CI fail is 😕

jenshalm:
The CI fail is caused by the tests for the sbt plugin. I'm using the built-in `match` command to compare file contents which is unfortunately not ignoring whitespace. If you revert all whitespace changes just under `sbt/src/sbt-test/site` the build should be green.

j-mie6:
Gah. Should be able to tackle this tonight, though my editor automatically trims whitespace, so getting it to comply will be a battle 😂

PR #416, opened 2023-05-23
Fix outdated code example.

zhongl:


jenshalm (approved):

jenshalm:
Thank you very much for spotting these. Unfortunately these code examples do not get checked by mdoc.

PR #443, opened 2023-06-29
remove unused parameters

jenshalm:
This removes all unused parameters (or replaces them with `_` in pattern matches).

Some signatures are public API, so these are breaking changes. But many of those APIs won't be public anymore anyway in M2, so it's a very minor point.

In two cases where removal is not possible I added a `@nowarn` annotation to the parameter.

Some of these improvements could be backported to 0.19, but I'd prefer to keep it simple here and aim for getting fatal warnings on for the main branch only.

There are 10 warnings left, which cannot be as easily grouped, so I might lump them together into a final PR and then try to switch on fatal warnings as a last step before a release.

Btw. I did not run a full code analysis on why the unused parameters existed in the first place and rely on the very large test suites instead. In most cases it will be historic leftovers that were not removed in a refactoring given that this is an 11-year-old code base.

### DIFF
--- a/core/shared/src/main/scala/laika/directive/api.scala
+++ b/core/shared/src/main/scala/laika/directive/api.scala
@@ -17,14 +17,15 @@
 package laika.directive
 
 import cats.{ Functor, Semigroupal }
-import laika.ast.{ TemplateSpan, _ }
-import laika.collection.TransitionalCollectionOps._
+import laika.ast.{ TemplateSpan, * }
+import laika.collection.TransitionalCollectionOps.*
 import laika.config.Origin.DirectiveScope
-import laika.config._
+import laika.config.*
 import laika.parse.SourceFragment
 import laika.parse.directive.DirectiveParsers.ParsedDirective
 import laika.parse.hocon.ConfigResolver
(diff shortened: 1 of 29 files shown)


core/shared/src/main/scala/laika/directive/api.scala
```diff
@@ -259,7 +260,7 @@ trait BuilderContext[E <: Element] {
         .asInstanceOf[E]
     }
 
-    def runsIn(phase: RewritePhase): Boolean = directive.fold(true)(_.runsIn(phase))
+    def runsIn(@nowarn phase: RewritePhase): Boolean = directive.fold(true)(_.runsIn(phase))
```
armanbilge, line 263: Another option here is to `import org.typelevel.scalaccompat.annotation.unused` (sbt-typelevel includes the necessary dependency by default). It's a bit more specific than a generic `@nowarn`.

```suggestion
    def runsIn(@unused phase: RewritePhase): Boolean = directive.fold(true)(_.runsIn(phase))
```

jenshalm: yup, that's nicer

PR #549, opened 2023-10-26
Update group ids after switch to typelevel org

sbrunk:
Since the 1.0 release, if you copy the maven coordinates from the docs, you'll get an invalid dependency.

I'm not completely sure about the scala-steward conf, but I'm assuming it should be changed too.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,7 +1,3 @@
-updates.ignore = [
-  { groupId = "org.planet42" }
-]
-
 pullRequests.frequency = "7 days"
 
 updates.limit = 3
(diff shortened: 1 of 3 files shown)


jenshalm:
Hello, and thanks for the fix.

The old scalasteward config existed as I preferred to have explicit control over the version of Laika I used for producing Laika's manual, but we cannot merge the change you did as it would then exclude all typelevel dependencies, e.g. cats.

I think it's best to simply remove the exclusion, everything else LGTM.

sbrunk:
Done! And thanks for creating and maintaining Laika :)

PR #596, opened 2024-03-11
Collect fragments while concatenating `xsl-fo`

nalkuatov:


jenshalm:
Many thanks for contributing. The implementation looks ok, but it needs a test before this can be merged. There is an `FOConcatenationSuite`, you can add the logic to the existing success case, or you can add a third test. I'm fine with both.

jenshalm:
Superseded by #607 which also contains a test and builds on #606 which properly deals with fragments with the same name.

PR #614, opened 2024-05-21
Preserve explicitly assigned header ids

### LINKED ISSUE #613: Manual assignment of header block ids is overwritten
The id for a `Header` block can be explicitly assigned with `withId`. However this id appears to be replaced at some point during the transformation pipeline with one generated using the theme's slug builder applied to the result of `extractText`.

I think the explicitly assigned id should be preserved.

milessabin:
Fixes #613.

Also fixes one test which relied on the buggy behaviour.


jenshalm:
Looks all good to me. Many thanks for finding and fixing this issue!

Issue #576, opened 2024-01-24
Theme documentation needs attention


noelwelsh:
The current documentation for custom themes could do with a rework.

There are two main problems:

1. The documentation doesn't give an overview of the process.
2. The documentation is completely backwards

After about 3 hours working on creating my own theme I think I need to create a `ThemeProvider` that builds a theme using `ThemeBuilder`. `ThemeBuilder` is only mentioned at the very end of the documentation, yet this is the type I should start with. The first type documented is `Theme`, which as best I can tell I will never explicitly define. This is a common problem of explaining all the intricate pieces, which are naturally interesting to the developer who created them, but doesn't address the problems the user has.

Better documentation would start with an overview: to create a custom theme create a `ThemeProvider` that builds a theme using a `ThemeBuilder`. Then go on to explain how these types work.

"While Laika offers the full API as for user inputs for maximum flexibility and consistency, it is recommended to avoid any file system resources by using addFile or addDirectory."

I think this sentence is saying I should *not* using `addFile` or `addDirectory`, but as written is it ambiguous. I think it reads better as "While Laika offers the full API as for user inputs for maximum flexibility and consistency, it is recommended to avoid using addFile or addDirectory to add any file system resources."

jenshalm:
Yes, I agree the structure is somewhat unfortunate. It was several years ago that I wrote this page, so I don't remember my motivation for doing it like this. I had to actually re-read it myself now before I could respond.

It seems to be that if the first section titled "The Theme Trait" is removed and instead replaced by a general overview of what type of functionality you usually assemble for a theme and what APIs you'll be using for this, then the other sections after that could probably remain unchanged.

I won't get to work on this before both 1.0.1 and 1.1.0 are out, so it'll take a bit, which is why I put it into Backlog for now.

PR #621, opened 2024-07-02
promote most internals of Helium's include directive to public API in laika.theme.*

jenshalm:
Previously the `@:includeJS` and `@:includeCSS` directives were only available within the Helium theme, since they cannot be configured via HOCON like most other directives which come bundled with `laika.core` (some of the API allows to pass a function argument which cannot be encoded in HOCON).

For this reason authors of alternative themes or end users not using Helium had to either hard-code the includes into their custom templates or re-build or copy a lot of the implementation details.

Since the directives themselves cannot be promoted to become core directives for the reasons stated above, this PR simply aims to remove about 90% of the boilerplate required to re-create them outside of Helium by promoting their builder APIs from an internal Helium package to the public `laika.theme.config` package.

Example for creating an instance of a `@:includeCSS` directive that includes all CSS documents in the `styles` directory:

```scala
val config = IncludeCSSConfig.empty.internalCSS(Root / "styles")
val includeDirective = IncludeDirective.forCSS(htmlConfig = config)
```

This instance can then be registered like any other directive:

```scala
object MyDirectives extends DirectiveRegistry {
  val spanDirectives = Seq()
  val blockDirectives = Seq()
  val templateDirectives = Seq(includeDirective)
  val linkDirectives = Seq()
}
```

After this step the `@:includeCSS` directive can be used in any template outside of Helium.

There are no new tests since Helium now delegate to the promoted API, meaning the directive is tested within the Helium suites.

Proper documentation will be added to the manual as part of #576.

FYI @noelwelsh

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -150,7 +150,8 @@ lazy val io = project.in(file("io"))
   .settings(
     name := "laika-io",
     libraryDependencies ++= Seq(catsEffect, fs2IO, munit, munitCE3),
-    Test / scalacOptions ~= disableMissingInterpolatorWarning
+    Test / scalacOptions ~= disableMissingInterpolatorWarning,
+    mimaBinaryIssueFilters ++= MimaFilters.includeRefactoring
   )
 
 lazy val pdf = project.in(file("pdf"))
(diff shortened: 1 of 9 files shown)


noelwelsh:
This looks great! It's going to save me a lot of work. Thanks!

Issue #706, opened 2025-04-25
Long page nav is permanently cut off


OddKristensen:
For pages with loads entries in the page nav it is impossible to scroll down to the bottom of the element. [http4s](https://http4s.org/changelog.html)'s changelog is a good example of this in action. Even if you scroll down to the bottom of the page, the page nav is still cut off because it has `position: fixed`.

Notice in the screenshot that even though I have scrolled to the bottom of the page, I still cannot see the last entries in the page nav.

![Image](https://github.com/user-attachments/assets/19506310-f0f8-469b-83b7-4472de023921)

jenshalm:
Fixed in #707

PR #707, opened 2025-04-25
Add scrollbar to #page-nav

OddKristensen:
Took a quick stab at [issue 706](https://github.com/typelevel/Laika/issues/706). There are many ways to solve this, but simply adding a height and a scrollbar on vertical overflow is the easiest solution that does not require additional changes to the DOM. Again, with [http4s](https://http4s.org/changelog.html)'s changelog as an example it will look like this:

Screenshot 2025-04-25 at 11 54 55

I picked `height: 80%` because it still looks somewhat ok an really short screens. I found that 90% was too cramped when the screen was getting really short.

With `height: 90%` it looks like the screenshot on the bottom, where you cannot actually scroll all the way down.

Screenshot 2025-04-25 at 12 01 18

Compared that wight `height: 80%`:

Screenshot 2025-04-25 at 11 55 33

### DIFF
--- a/io/src/main/resources/laika/helium/css/nav.css
+++ b/io/src/main/resources/laika/helium/css/nav.css
@@ -346,7 +346,7 @@ ul.nav-list, #page-nav ul {
 #page-nav {
   display: none;
   position: fixed;
-  height: 80%;
+  max-height: calc(100% - 90px - var(--top-bar-height));
   overflow-y: auto;
   top: calc(45px + var(--top-bar-height));
   right: 45px;
--- a/io/src/main/resources/laika/helium/css/nav.css
+++ b/io/src/main/resources/laika/helium/css/nav.css
@@ -346,6 +346,8 @@ ul.nav-list, #page-nav ul {
 #page-nav {
   display: none;
   position: fixed;
+  height: 80%;
+  overflow-y: auto;
   top: calc(45px + var(--top-bar-height));
   right: 45px;
   width: 250px;


io/src/main/resources/laika/helium/css/nav.css
```diff
@@ -346,6 +346,8 @@ ul.nav-list, #page-nav ul {
 #page-nav {
   display: none;
   position: fixed;
+  height: 80%;
```
jenshalm, line 349:
```suggestion
  max-height: calc(100% - 90px - var(--top-bar-height));
```

The problem with the suggested solution is that it fixes the edge cases while breaking the most common cases: all pages with just a few entries in page navigation now get a long box with a lot of empty space which looks broken.

Using `max-height` instead of `height` fixes part of the problem. But it still keeps the issue you observed where it can start to look strange with very small page heights, because it does not preserve a fixed margin below the navigation box. The suggested change also improves on that: by using 100% and subtracting the height of the top navigation and 90 pixel for the space above and below the navigation box, we force a fixed margin on the bottom.

Would be great if you'd also quickly test the suggested change before accepting it.

jenshalm (changes requested):
Welcome, and thanks for preparing this PR. I wasn't aware of this issue.

Please verify the one suggested change, as the proposed solution breaks pages with short page navigation boxes.

io/src/main/resources/laika/helium/css/nav.css
OddKristensen: Thank you for the refinement, those look great!

jenshalm (approved):

jenshalm:
Closes #706

PR #754, opened 2026-02-28
fix rendering of 404 page for typelevel website

### LINKED ISSUE #602: Rendering of the 404 page is broken
https://typelevel.org/doesnt/exist

The problem is that Laika generates the HTML for 404.md using relative paths, but the page may be rendered from any path. This may be something that we need to fix upstream in Laika.

abby-ql:
Fixes typelevel/typelevel.github.com#602

the typelevel website previously would break for https://typelevel.org/doesnt/exist because laika renders internal targets in relative path. The relative `main.css` resolve to `/doesnt/main.css` thus assets failed to load. I think the issue is in `HTMLRenderer.renderTarget` where internal targets are rendered relative to `fmt.path`. My fix is basically

-detect when rendering the 404 page (`fmt.path.basename == "404"`)
-render internal targets as root-relative (`/main.css`) for the 404 page

I have also done testing & checked
-the generated `404.html` now contains `/main.css` instead of `main.css` for static site
-`http://127.0.0.1:8080/doesnt/exist` worked when served locally
-other pages (like `index.html`) unaffected and still use relative path

### DIFF
--- a/core/shared/src/main/scala/laika/api/Renderer.scala
+++ b/core/shared/src/main/scala/laika/api/Renderer.scala
@@ -136,23 +136,28 @@ abstract class Renderer private[laika] (val config: OperationConfig, skipRewrite
         .leftMap(InvalidConfig(_))
     }
 
-    (if (skipRewrite) Right(targetElement) else rewrite).map { elementToRender =>
-      val renderContext =
-        new Formatter.Context[Formatter](
-          renderFunction,
-          elementToRender,
-          Nil,
-          styles,
-          doc.path,
-          pathTranslator,
-          if (config.compactRendering) Indentation.none else Indentation.default,
-          config.messageFilters.render
-        )
+    (if (skipRewrite) Right(targetElement) else rewrite).flatMap { elementToRender =>
+      doc.config
+        .getOpt[String]("laika.renderTarget.absolute.baseUrl")
+        .leftMap(InvalidConfig(_))
+        .map { baseUrlOpt =>
+          val renderContext =
+            new Formatter.Context[Formatter](
+              renderFunction,
+              elementToRender,
+              Nil,
+              styles,
+              doc.path,
+              pathTranslator,
+              if (config.compactRendering) Indentation.none else Indentation.default,
+              config.messageFilters.render,
+              baseUrlOpt
+            )
 
-      val formatter = format.formatterFactory(renderContext)
-
-      renderFunction(formatter, elementToRender)
-    }
+          val formatter = format.formatterFactory(renderContext)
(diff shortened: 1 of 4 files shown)


armanbilge (changes requested):
Thanks for the PR! I think this is a step in the right direction.

However, my concern is that this solution is too bespoke.

A different idea I have to solve this is by introducing a setting to the per-document HOCON config. For example, something like:

```hocon
laika.renderTarget.absolute.baseUrl = "/"
```

(We can bikeshed the exact config name).

If a document contains this config, then all relative targets in that document should be rendered as absolute, with the specified base URL.

---

Alternatively, we could introduce a Laika Extension that is specific to GitHub Pages, that uses this bespoke implementation.

abby-ql:
Thank you so much for the feedback! @armanbilge I rewrote & added a per-document configuration option
`laika.renderTarget.absolute.baseUrl = "/"`
When present in a document’s HOCON config, all internal targets in that document are rendered as absolute url using the specified base url. changes:

-wrapping the PathTranslator for that document specifically added new `AbsoluteBasePathTranslator.scala`
-converting resolved InternalTarget instances into ExternalTarget instances with the configured base path

I also added a test in `TreeRendererSpec` that checks
-docs with laika.renderTarget.absolute.baseUrl render internal links as absolute URLs
-docs without the setting continue to render links relative
I have done similar checks as before to verify the relative path did changed to `/main.css etc` in 404.html generated

armanbilge (changes requested):
Thanks for considering the feedback!

Do you think there is a way we could use Laika's existing `internalTargetsAbsolute` feature somehow?

https://github.com/typelevel/Laika/blob/2f7483f5b0d11715cfca811d122a3136e740238a/core/shared/src/main/scala/laika/ast/Target.scala#L120-L122

abby-ql:
Thanks for the feedback! Indeed this can be done in simpler ways without wrapping and introducing new things. I changed the implementation to read the configured base Url in Renderer and pass it through `Formatter.Context` and also now it uses `Target.render(internalTargetsAbsolute = true)` , and updated the tests too.

reardonj:
I was trying to take a look at this. I am a bit perplexed at what is going on here: 
https://github.com/typelevel/Laika/blob/b87fc9906bdad8200505fd26245bde5205ef0c42/core/shared/src/main/scala/laika/ast/Target.scala#L103-L111

The documentation for this `Resolved` class seems to suggest there is already a concept of a base URL, though I haven't hunted down where that is used. I am wondering if it is possible to do this by overriding that for a single page instead of all the extra logic here. I'm not really familiar enough with this codebase to properly answer the question though.

PR #757, opened 2026-03-08
Add dark mode toggle

### LINKED ISSUE #382: Dark Mode Toggle
It would be nice if a toggle button for dark mode could be supported for HTML. As I understand it, the use of dark mode or not is dependent on your operating system or browser configuration, which works great, but it would also be nice if the user can toggle it on and off on a per-case basis: just like in scala3doc!

abby-ql:
Closes #382
Helium currently supports dark mode but users cannot manually override it per site.
This PR adds a manual color mode toggle to the Helium theme. Users can switch between `Auto` (follow `prefers-color-scheme`), `Light`, and `Dark`. The selected mode is stored locally in the browser using `localStorage` and persists across page navigations. If no selection is made,  the default is `Auto` which follows the system/browser preference, and no personal or identifying data is stored. The toggle button is placed on top left navigation bar.
Testing: checked `Auto` does follow browser default, and Light/Dark mode switch is working.

Example ( using the protosearch doc) 
image
image

### DIFF
--- a/io/src/main/resources/laika/helium/css/nav.css
+++ b/io/src/main/resources/laika/helium/css/nav.css
@@ -25,6 +25,29 @@ header a:hover, nav .row a:hover {
   color: var(--component-hover)
 }
 
+header .row button.theme-toggle {
+  margin: 0 0 0 20px;
+  padding: 0;
+  border: none;
+  background: none;
+  color: var(--component-color);
+  cursor: pointer;
+  font: inherit;
+  font-size: 0.9rem;
+  opacity: 0.8;
+}
+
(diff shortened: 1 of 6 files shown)


abby-ql:
The current implementation renders the toggle (and the initialization script) by default, which changes the generated html structure and causes several Helium html tests to fail due to expected output differences. I have checked functionally everything works as intended in the browser. 

Would like to check whether enabling the toggle by default ( which would require updating a whole bunch of tests) is preferred, or if it should instead be opt in via Helium configuration flag, thank you!

PR #774, opened 2026-06-25
Cross build laika-sbt for sbt 2

### LINKED ISSUE #773: sbt 2.x support for `laika-sbt`
`sbt-typelevel` depends on `laika-sbt` https://github.com/typelevel/sbt-typelevel/issues/899

---

I'm working on cross-building for sbt 2.x, and I found sbt 2.0.0 has a macro crash when a `Def.setting` or `Def.task` body contains a generic method call with a HKT type argument. Laika hits this through its `IO`-based builder APIs, so sbt 2 support may need an upstream sbt fix https://github.com/sbt/sbt/issues/9375

tanishiking:
Depends on upstream sbt change https://github.com/sbt/sbt/pull/9377, ~we need to wait until sbt 2.0.1~ 2.0.1 is released with that fix.

---

close https://github.com/typelevel/Laika/issues/773

- Exclude all laika-sbt tasks from sbt 2 caching with `Def.uncached`.
- Adapt to sbt2 API changes using https://github.com/sbt/sbt2-compat if you don't like adding dependency, we can define small set of PluginCompat.
- Run the plugin build and tests in CI with sbt 2, Scala 3, and Java 17.
- Set `Laika / target` explicitly in each scripted project `build.sbt` so scripted tests work across sbt 1 and 2 target directory.

Depends on upstream sbt change https://github.com/sbt/sbt/pull/9377


sbt/src/main/scala/laika/sbt/LaikaConfig.scala
```diff
@@ -182,6 +182,4 @@ object LaikaConfig {
     logLevel = MessageFilter.Warning
   )
 
-  private def unapply(conf: LaikaConfig) = conf
```
tanishiking, line 185: unused, removed for suppress warning in Scala3.

sbt/src/main/scala/laika/sbt/LaikaPlugin.scala
```diff
@@ -155,39 +156,47 @@ object LaikaPlugin extends AutoPlugin {
...
-    laikaXSLFO                  := Tasks.generate.toTask(" xslfo").value,
-    laikaEPUB                   := Tasks.generate.toTask(" epub").value,
-    laikaPDF                    := Tasks.generate.toTask(" pdf").value,
-    laikaAST                    := Tasks.generate.toTask(" ast").value,
-    laikaPackageSite            := Tasks.packageSite.value,
-    laikaPreview                := Tasks.startPreviewServer.value,
-    Laika / clean               := Tasks.clean.value,
-    laikaSite / mappings        := Def.sequential(Tasks.site, Tasks.mappings).value
-  ) :+ (cleanFiles += (Laika / target).value)
+  @nowarn("msg=unused import")
```
tanishiking, line 159: For `sbtcompat.PluginCompat.*` in Scala3.

sbt/src/sbt-test/site/xslfo/build.sbt
```diff
@@ -5,3 +5,5 @@ version := "0.1"
 scalaVersion := "2.12.6"
 
 enablePlugins(LaikaPlugin)
+
+Laika / target := baseDirectory.value / "target" / "docs"
```
reardonj, line 9: Did this have to be added due to sbt2 changes the target directory setup?

build.sbt
```diff
@@ -188,22 +198,56 @@ lazy val plugin = project.in(file("sbt"))
...
+      }
+    },
+    // sbt 2.x is built with Scala 3.8, which requires Java 17 classfiles.
+    // https://github.com/typelevel/sbt-typelevel/issues/851
+    tlJdkRelease                  := {
+      if (scalaBinaryVersion.value == "3") Some(17)
+      else Some(8)
+    },
+    // Scaladoc 3 cannot currently resolve some TASTy from sbt 2 transitive dependencies.
+    // https://github.com/scala/scala3/issues/18487
```
reardonj: I see you fixed this bug. I guess we are stuck until at least 3.9 to make use of it though?

build.sbt
```diff
@@ -188,22 +198,56 @@ lazy val plugin = project.in(file("sbt"))
...
+    scripted                      := scripted.dependsOn(
       core.jvm / publishLocal,
       io / publishLocal,
       pdf / publishLocal,
       preview / publishLocal
     ).evaluated,
+    // TODO: Re-enable MiMa for sbt 2 once a previous sbt 2 plugin artifact exists.
+    // The sbt 2 / Scala 3 plugin artifact did not exist in previous Laika releases,
+    // so MiMa cannot compare against a missing baseline.
+    mimaPreviousArtifacts         := {
```
reardonj: I don't suppose there is a nicer way to do this @armanbilge ?

sbt/src/main/scala/laika/sbt/LaikaPlugin.scala
```diff
@@ -155,39 +156,47 @@ object LaikaPlugin extends AutoPlugin {
...
+  override def projectSettings: Seq[Setting[?]] = {
+    import sbtcompat.PluginCompat.*
+
+    Seq(
+      Laika / sourceDirectories   := Seq(sourceDirectory.value / "docs"),
+      Laika / excludeFilter       := HiddenFileFilter,
+      laikaInputs                 := Def.uncached(Settings.defaultInputs.value),
```
reardonj: Do these need to be uncached or is this just for ease of migration?

reardonj (commented):
This looks pretty reasonable. I'm not familiar with the CI setup stuff enough to comment on it.

build.sbt
tanishiking: When I was investigating the bug, I realized that it matters only when the plugin has a dependency to any symbols that is annotated with `@datatype.data` (such as https://github.com/sbt/sbt/blob/d4e8fb583c1a61733500112a2ec6dd8af2a4de4c/lm-coursier/definitions/src/main/scala/lmcoursier/CoursierConfiguration.scala#L23).

So there's no problem for `laika`, I'll remove this workaround 🙇 

done https://github.com/typelevel/Laika/pull/774/commits/5c9a31790e11f35f7aa9279ae0329f05b3e42a79

sbt/src/sbt-test/site/xslfo/build.sbt
tanishiking: Yes, by default, it's `target/out/jvm/scala-xxx/<moduleName>` for sbt 2.x, while sbt 1.x is `target/...`. If we didn't add this configuration, test script's `$ must-mirror target/docs/site/hello.html` doesn't work across sbt 1.x and 2.x.

Alternatively, we can define a custom sbt task that checks the content of generated files, but I think configuring `Laika / target` would be the easiest fix (smallest diff).

sbt/src/main/scala/laika/sbt/LaikaPlugin.scala
tanishiking: Oops, nice catch, they're `settingKey` (not `taskKey`) and we didn't need `Def.uncached`. (https://github.com/typelevel/Laika/pull/774/commits/5c9a31790e11f35f7aa9279ae0329f05b3e42a79)

build.sbt
tanishiking: Oh, I learnt that there's `tlVersionIntroduced` in `sbt-typelevel`:

https://github.com/typelevel/sbt-typelevel/blob/dfb6b58df7bb55d8ac8b65658a0f4e5d6a4ca445/docs/faq.md?plain=1#L30-L34

https://github.com/typelevel/sbt-typelevel/blob/dfb6b58df7bb55d8ac8b65658a0f4e5d6a4ca445/docs/faq.md?plain=1#L30-L34

Maybe we wanna set `tlVersionIntroduced := Map("3" -> "1.3.3")`  (where 1.3.3 is the next release version, maybe `1.4.0` ?) 
Then we don't need to manually remove this `mimaPreviousArtifacts` when we have scala3 release.

tanishiking: Set `tlVersionIntroduced := Map("3" -> "1.3.3")`. I'm not sure the next version is either `1.3.3` or `1.4.0` though :) 
https://github.com/typelevel/Laika/pull/774/commits/e4dae94272e830dde1efc916c0ce2233ea5b203c

sbt/src/sbt-test/site/xslfo/build.sbt
reardonj: I think it's fine as is. Just was confirming my understanding.

build.sbt
reardonj: I believe a new target would get a minor release, so it should be 1.4.0

tanishiking: 👍 updated to 1.4.0 https://github.com/typelevel/Laika/pull/774/commits/93267131ade58ba9a567110334a4f12200f4a8e5

reardonj:
FYI, force pushing makes it basically impossible to review what actually changed since you destroy the history when do this. I’d have to review this whole PR from scratch now because all the commits changed so I can’t assume there are no other changes.

tanishiking:
Ah, sorry I should've used fixup commits

sbt/src/main/scala/laika/sbt/Tasks.scala
```diff
@@ -323,8 +329,9 @@ object Tasks {
     generate.toTask(formats.mkString(" ", " ", ""))
   }
 
-  val mappings: Initialize[Task[Seq[(File, String)]]] = task {
-    sbt.Path.allSubpaths((laikaSite / target).value).toSeq
+  val mappings: Initialize[Task[Seq[(FileRef, String)]]] = task {
+    implicit val fileConverter: FileConverter = (ThisBuild / sbt.Keys.fileConverter).value
```
j-mie6, line 333: Not `given FileConverter`?

sbt/src/main/scala/laika/sbt/Tasks.scala
```diff
@@ -378,7 +385,14 @@ object Tasks {
     val zipFile      = (Laika / target).value / artifactName
     streams.value.log.info(s"Packaging $zipFile ...")
 
-    sbt.IO.zip((laikaSite / sbt.Keys.mappings).value, zipFile, None)
+    implicit val fileConverter: FileConverter = (ThisBuild / sbt.Keys.fileConverter).value
```
j-mie6, line 388: ditto

sbt/src/sbt-test/site/messageFilters/build.sbt
```diff
@@ -8,5 +8,7 @@ scalaVersion := "2.12.6"
 
 enablePlugins(LaikaPlugin)
 
+Laika / target := baseDirectory.value / "target" / "docs"
```
j-mie6, line 11: Is this something that will change in the downstream user configuration? Will we need to document these changes somewhere?

reardonj: If I understand correctly, this is just a test file that is being forced to use an sbt 1.x output path for testing convenience. Actual users of Laika in sbt 2 would see their output follow standard sbt 2 behaviour. If the user upgrades and has other script assume the output path, they would see issues, so it would be good to note the behaviour as part of the cross-build release.

sbt/src/main/scala/laika/sbt/Tasks.scala
tanishiking: Unfortunately, since we're cross-building against sbt1 and sbt2, and we can't use `given` here IIUC.

sbt/src/sbt-test/site/messageFilters/build.sbt
tanishiking: > f I understand correctly, this is just a test file that is being forced to use an sbt 1.x output path for testing convenience.

Yes, that's right :+1:
https://github.com/typelevel/Laika/pull/774#discussion_r3636430341

custommonkey:
PR #783 should fix the Java version that's used by sbt-typelevel which is causing the Scala Steward validation to fail.

tanishiking:
@custommonkey Thanks! Update `sbt-typelevel` to 0.8.7 fixed the problem thanks to https://github.com/typelevel/sbt-typelevel/pull/905 :+1:

build.sbt
```diff
@@ -24,14 +24,24 @@ inThisBuild(
     scalaVersion       := versions.scala2_12,
     developers         := List(Developer("jenshalm", "Jens Halm", "", url("http://planet42.org"))),
     tlCiHeaderCheck    := false,
-    tlCiDependencyGraphJob := false,
-    githubWorkflowJavaVersions += JavaSpec.temurin("17"),
+    tlCiDependencyGraphJob     := false,
+    githubWorkflowJavaVersions := Seq(JavaSpec.temurin("17"), JavaSpec.temurin("8")),
```
zainab-ali: This doesn't generate the correct build matrix. From [the GitHub Actions run](https://github.com/typelevel/Laika/pull/774/checks
), it looks like `rootJVM` isn't compiled on Scala 3 at all.

I think we want to keep the same matrix, but build the plugin module for Scala 3, JDK 17. The following exclusions might work better:

```scala
    githubWorkflowJavaVersions += JavaSpec.temurin("17"),
    githubWorkflowBuildMatrixExclusions := {
      val java17 = JavaSpec.temurin("17").render
      githubWorkflowBuildMatrixExclusions.value.filterNot(
        _.matching == Map("scala" -> "3", "java" -> java17)
      ) ++ List(
        MatrixExclude(Map("project" -> "plugin", "scala" -> "2.12", "java" -> java17)),
        MatrixExclude(Map("project" -> "rootJVM", "scala" -> "3", "java" -> java17)),
        MatrixExclude(Map("project" -> "plugin", "scala" -> "2.13"))
      )
    },

```

build.sbt
```diff
@@ -188,22 +198,42 @@ lazy val plugin = project.in(file("sbt"))
...
+        case "2.12" => versions.sbt1
+        case _      => versions.sbt2
+      }
+    },
+    scriptedSbt                   := {
+      scalaBinaryVersion.value match {
+        case "2.12" => versions.sbt1
+        case _      => versions.sbt2
+      }
+    },
```
zainab-ali: The default for `scriptedSbt` is the `pluginCrossBuild / sbtVersion` value, so this setting is redundant.

```suggestion
```

sbt/src/main/scala/laika/sbt/LaikaPlugin.scala
zainab-ali: Since this only occurs on Scala 3, you can use [scalac-compat](github.com/typelevel/scalac-compat). This is already a provided library dependency.

```suggestion
  @org.typelevel.scalaccompat.annotation.nowarn3("msg=unused import")
```

zainab-ali (commented):
The `laikaSite` and `laikaPreview` tasks work for me locally in [sbt-typelevel](https://github.com/typelevel/sbt-typelevel/compare/main...zainab-ali:cross-build-all-plugins) on Scala 3 🎉

build.sbt
tanishiking: Ah, right, done in https://github.com/typelevel/Laika/pull/774/commits/1b5139893af6db08b4ce1539b5c23aef4a8beea4. And your suggestion looks much better, thanks!

On top of that, 
- Excluded `plugin + scala 3 + java 8`  https://github.com/typelevel/Laika/pull/774/commits/f4497d49b1196c1d969751d169be00f8a32c6a63 because sbt 2 requires JDK >= 17
- Don't exclude `rootJVM + sccala 3 + java 17` https://github.com/typelevel/Laika/pull/774/commits/dccc83a83ac3c48dd2ccd69a40e46b59cc9ae035 because I don't feel there's reason to exclude?

